PR #1, opened 2013-04-15
Fixed bug 10833 - wrong text in revAbout (community edition says commerc...

runrevmark:
...ial).


runrevmark:
Merged into release-6.0.1

PR #2, opened 2013-04-15
Fixed bug 10829 - corrected iOS device binary references in installer pa...

runrevmark:
...ckage script.


runrevmark:
Merged into release-6.0.1

PR #4, opened 2013-04-16
Installer package updates to support building both Community and Commerc...

runrevmark:
...ial editions.


runrevmark:
Merged into release-6.0.1

PR #5, opened 2014-01-04
allow 'go to definition' in system stacks

mwieder:
This is a fix for bug 10722 and involves commenting one line of code, as described in the bug report. This allows 'go to definition' to work in system stacks, which allows developers to more easily understand what's going on in their (and the IDE's) scripts. It also allows bugs to be tracked and fixed in system stacks. I can see no downsides to having this happen.

I _am_, however, a bit concerned that the original 6.5.1 stack is ~3MB in size and after I commented one line and saved the stack it compressed to under 2MB. It still works fine, but I'm at a loss to explain the difference.


livecodeali:
Hi Mark,

Thanks for this, I've reimplemented it against develop-6.7 now that we have our git IDE processes worked out a bit better, so I'm closing this pull request. I just deleted the whole clause that starts 'if not gRevDevelopment...'

https://github.com/runrev/livecode-ide/pull/201

I've scriptified the script editor behaviors in the 8.0 ide so will make a separate pull request for that one.

PR #9, opened 2014-11-07
bugfix 13953

angerangel:
This fix bug 13953, http://quality.runrev.com/show_bug.cgi?id=13953


livecodeali:
Hi MaxV,

Unfortunately we cannot accept outside contributions to binary stack files, as without being able to see the diff we can't review the changes. We can only do that with 'script only' stacks.

If you would like to describe the exact changes in your fix we can implement them here.

angerangel:
Don't you have a livecode diff viewer? Are you serious?
However I corrected the path in linux environment, it pointed to revSpecialFolderPath("documents") but that path doesn't work on linux.
P.S. do you need a livecode diff viewer?

angerangel:
The code is in Stack and is the following change:

```
on preOpenStack
   local tDatabaseFile
   #check specialFolderPath("documents") is not empty
   if specialFolderPath("documents") is not empty then      
   ## Connect to the SQLite database using the database file path
      put specialFolderPath("documents") & "/bookLibrary.sqlite" into tDatabaseFile
   else
      put specialFolderPath("temporary") & "/bookLibrary.sqlite" into tDatabaseFile
   end if
```

livecodeali:
The problem is that we don't have something that can show the complete diff for two binary stackfiles. It would be possible to hide malicious code in the scripts of various other objects, or alter the value of an important custom property, or move objects around or delete them, or add unwanted objects etc. I'm not saying that you're at all likely to have done any of those things!

livecodeali:
Implemented here:
https://github.com/runrev/livecode-ide/pull/211

PR #32, opened 2015-02-20
[[ Developer Extensions ]] Locked: Plugins/revExtensionBuilder.livecode

livecodeali:
Tweak library and plugin to ensure path to LCB file is displayed in the extension selection menu even if it doesn't compile.


livecodeali:
Superseded by https://github.com/runrev/livecode-ide/pull/33

PR #35, opened 2015-02-24
[[ Message Box ]] Updated message box and revidelibrary

livecodegeorgia:
-- lock screens for when header/footer size is changed
-- updated revidelibrary to include functions for message box


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1383,10 +1383,42 @@ end revIDEGlobalVariables
 
 function revIDEGlobalProperties
    set the itemdel to slash
-   set the defaultfolder to item 1 to -3 of the filename of me
+   set the defaultfolder to item 1 to -4 of the filename of me
    return URL("file:globalProperties.txt")
```
livecodeali, line 1386: You should use `revIDESpecialFolderPath("IDE")` here instead of `item 1 to -4 of the filename of me`` because that calculates relative to stack "home" which won't move. Plus you won't need to set the itemdelimiter :-)

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1383,10 +1383,42 @@ end revIDEGlobalVariables
...
 end revIDEGlobalProperties
 
+function revIDEGlobalPropertyRange pProperty
+   local tRangeArray, tLine
+   
+   set the itemdel to slash
+   set the defaultfolder to item 1 to -4 of the filename of me
```
livecodeali: Similarly here.

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1383,10 +1383,42 @@ end revIDEGlobalVariables
...
+   end repeat
+   
+   return revRangeArray[pProperty]
+end revIDEGlobalPropertyRange
+
+function revIDEGlobalPropertyContinuousRange pProperty
+   local tContinuousArray, tLine
+   
+   set the itemdel to slash
+   set the defaultfolder to item 1 to -4 of the filename of me
```
livecodeali: And here.

PR #41, opened 2015-02-28
[[ Build Widgets ]] Update extension library to fetch packages from _build folder

livecodeali:


runrevmark:
Reviewed-by: @runrevmark.

PR #52, opened 2015-03-04
[[ Property Definitions ]] Tweak some property definitions to prevent errors

livecodeali:


runrevben:
Reviewed. The change in the entire file is due to line ending update.

PR #54, opened 2015-03-04
[[ bug 14634 ]] can't open livecodescript files from file>open menu

runrevben:
I've update the file chooser to allow the selected of "livecodescript" extension.

A script only is invisible by default so opening a livecodescript file appears to do nothing. I'm not sure if this is the correct behavior but making it visible doesn't seem correct. Only expert users will use script only's at this stage so would be able to edit the script via the message box. If there is a suggestion of a different behavior very happy to discuss and implement a fuller solution that is clearer for newer users.

### DIFF
--- a/Toolset/libraries/revcommonlibrary.livecodescript
+++ b/Toolset/libraries/revcommonlibrary.livecodescript
@@ -1385,7 +1385,7 @@ end revAnswerFile
 function revAnswerFiles pType, pPrompt
   switch pType
   case "stack"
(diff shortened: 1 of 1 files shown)


livecodestephen:
Would it make sense to immediately open the script in the script editor when opening a script only stack?

runrevmark:
@livecodestephen: I think that does make a lot of sense - if you stack is script-only the only thing you can do with it is edit it and why would you be opening one from the file menu if you weren't intending to edit it?

livecodeali:
Filed an enhancement request. 
http://quality.runrev.com/show_bug.cgi?id=14831

PR #57, opened 2015-03-04
[[ Bug 14744 ]] Regenerate docs after fixing bug 14744

livecodeali:


runrevben:
Not something you can really review as such. But regenerating the docs is a sensible thing to do.

PR #58, opened 2015-03-04
[[ Docs Integration ]] A number of changes related to integrating docs.

livecodeali:
- Fixing to the home stack to make it compile in strict mode
- Using the correct mechanism to load the docs library
- Relocating useful functions (revIDELastModifiedTimeOfFile and revIDEExtensionFolders) to the IDE library
- Tweaking the LC Script generation code so that (theoretically) anyone can build the docs provided they have checked out the livecode-docs repo
- Reworking the API and guide data generation so that it a) only regenerates api json if the api.js file is out of date, and b) uses the extension library to provide relevant data
- Tweaking the docs library to use the passed in data for dictionary display, rather than using a default (eg use widget display name instead of type id)

### DIFF
--- a/Toolset/home.livecodescript
+++ b/Toolset/home.livecodescript
@@ -1012,9 +1012,8 @@ command revInternal__InitialiseLibraries
    end try 	 	
    	    	 	
    try 	 	
-      insert script of stack "revidedocumentationlibrary" into back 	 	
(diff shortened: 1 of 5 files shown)


runrevben:
Looks good. The 5th commit is a little difficult the fully review from a functional point of view but from a coding standpoint nothing stands out as problematic.

PR #66, opened 2015-03-06
[[ Bug 14775 ]] LOCKED: revaskdialog.rev, revanswerdialog.rev. Icon buttons and fields should be transparent on ask and answer dialogs

livecodeali:


livecodeali:
Just changed the opaque property of 2 controls on the answer dialog stack, and 1 on the ask dialog stack.

PR #67, opened 2015-03-06
[[ Property Definitions ]] Temporarily remove accelerator modifiers group from property inspector

livecodeali:
Will reinstate once we have designed and built a reasonable editor widget/group.


livecodefraser:
Reviewed-By: Fraser Gordon

PR #71, opened 2015-03-06
new default extension icons

runrevben:


livecodeali:
These should do nicely.

PR #80, opened 2015-03-08
[[ Extensions ]] Return empty from __extensionError

livecodeali:
Otherwise 'the result' from a different operation makes its way into the extension properties array.


livecodefraser:
Reviewed-By: Fraser Gordon

PR #81, opened 2015-03-09
[[ Scriptify ]] LOCKED: revMessageBox.8, revCore.8, revDictionary.8, revIDEProjectBrowserView; Scriptify the palette behavior script, and remove button from revCore

livecodeali:


livecodeali:
This causes problems for the IDE at the moment. I am currently unsure why.

livecodeali:
The problem was that without a behavior set on it, the message box script wouldn't compile and therefore wouldn't load the behavior.

I have changed the behavior saved in the stackfile to the new scriptified version.

PR #84, opened 2015-03-09
[[ Bug ]] Linux seems to want fontawesome to be loaded second

livecodeali:
Tracked linux fontawesome issue down to the additional load of lcideicons.ttf - there may be a problem with loading multiple fonts dynamically on linux. This is merely a workaround for DP 1.


livecodefraser:
Reviewed-By: Fraser Gordon

Do the icons from the IDE font disappear as a result of this?

livecodeali:
Weirdly, it seems they do not, unless I'm looking in the wrong place.

PR #87, opened 2015-03-10
[[ Property Metadata ]] Parse property metadata generically into array

livecodeali:


livecodefraser:
Reviewed-By: Fraser Gordon

PR #97, opened 2015-03-11
[[ Bug 14878 ]] LOCKED: revMenuBar.rev, home.rev Move group 'Restore Development Tools' from defunct revLibrary stack onto revMenuBar stack

livecodeali:


livecodeali:
Since we will probably be looking at scriptifying revMenuBar in the future, I have taken the opportunity to make revRestore a stack in its own right. It now contains the replacement menuBar for when development tools have been suspended. Home stack's stack mapping list updated accordingly.

PR #103, opened 2015-03-11
[[ Bug ]] Allow all docs to be inline at the module description level

livecodeali:


livecodefraser:
Reviewed-By: Fraser Gordon

PR #109, opened 2015-03-12
[[ Bug ]] Add number editor for integer / real lcb properties

livecodeali:


livecodefraser:
Reviewed-By: Fraser Gordon

PR #112, opened 2015-03-12
[[ Tools Palette ]] Use object titles for tooltips

livecodeali:


livecodestephen:
I suppose this fixes [Bug 14894](http://quality.runrev.com/show_bug.cgi?id=14894)

PR #125, opened 2015-03-16
[Bug 14968] Always show errors from message box evaluation.

peter-b:
The message box should always show errors in its echo area,
regardless of the setting of the IDEDebugMode preference.


livecodeali:
Reviewed.

PR #126, opened 2015-03-17
Add various file extensions to list to be treated as text

livecodeali:


runrevben:
do we want to have "mlc" is here if we are trying to switch everything to "lcb"?

PR #128, opened 2015-03-18
[[ Widget Properties ]] Add kind, textFont, textSize and textAlign properties to widget property definitions

livecodeali:


Toolset/resources/supporting_files/property_definitions/propertyInfo.txt
```diff
@@ -73,6 +73,7 @@ id	ID	Basic	com.livecode.pi.image_id	true	false
 ink	Ink	Colors	com.livecode.pi.enum	true	false		srcCopy	blendClear,blendSrc,blendDst,blendSrcOver,blendDstOver,blendSrcIn,blendDstIn,blendSrcOut,blendDstOut,blendSrcAtop,blendDstAtop,blendXor,blendPlus,blendMultiply,blendScreen,blendOverlay,blendDarken,blendLighten,blendDodge,blendBurn,blendHardLight,blendSoftLight,blendDifference,blendExclusion,clear,noop,notSrcAnd,notSrcAndReverse,notSrcCopy,notSrcOr,notSrcOrReverse,notSrcXor,reverse,set,srcAnd,srcAndReverse,srcCopy,srcOr,srcOrReverse,srcXor,blend,addPin,addOver,subPin,transparent,adMin,addMax
 innerGlow	Inner Glow	Effects	com.livecode.pi.graphicEffect	true	false			
 innerShadow	Inner Shadow	Effects	com.livecode.pi.graphicEffect	true	false			
+kind	Kind	Basic	com.livecode.pi.string	true	true		no_default
```
runrevmark, line 76: Kind is read-only at the moment, and even it becomes settable it will only be for very specific purposes and for a short while... It probably shouldn't be in the PI list?

livecodeali, line 76: Read only is set on it here (the 6th item) - I thought it might be useful to be able to see it in the PI for the moment.

PR #148, opened 2015-03-25
[[ Bug ]] Ensure widgets are loaded in dependency order

livecodeali:


Toolset/libraries/revideextensionlibrary.livecodescript
```diff
@@ -668,6 +668,68 @@ private function __fetchModuleData pExtensionFile
    return tDataA
 end __fetchModuleData
 
+private function __extensionIsBuiltin pID
+   set the itemdelimiter to "."
+   return the number of items in pID is 3
+end __extensionIsBuiltin
+
```
runrevben, line 675: Is this viable when custom domain registration such as com.devore.widget are allowed?

livecodeali, line 675: No, it isn't - I wasn't sure exactly what the conditions were / will be which was why I made it a separate function so we can easily change it.

runrevben, line 675: Sounds sensible. Is this something we'll need to add to the IDE when it loads widgets so it tracks where the widget was loaded from? I'm happy to review it for now but we should add this to the todo so we don't forget. Thoughts?

runrevben:
Reviewed in the basis that Ali has added not for us to address identification of widgets when manifest updates are in place.

PR #149, opened 2015-03-25
Use segmented control for text align and text style properties

livecodeali:


runrevben:
Text changes look good. Assuming binary changes are good to merge so marking as reviewed.

PR #152, opened 2015-03-26
[[ Bug 14892 ]] Objects flicker when dragging from tools palette

livecodeali:


runrevben:
Excellent. Should have thought about this one before! I suspect that the changes in 6.7 with cocoa has caused this change in behaviour so that this code no longer worked. It was the way the old IDE did the creation of the drag element. Nice job looking into it.

PR #177, opened 2015-04-17
Feature analytics

livecodeali:


Toolset/libraries/revanalyticslibrary.livecodescript
```diff
@@ -0,0 +1,125 @@
...
+            end if
+         end if
+         _internal deploy windows tDeployInfo
+         break
+      case "Linux"
+         if revEnvironmentIsInstalled() then
+            put it & "/Linux/x86-64/Standalone" into tDeployInfo["engine"]
+         else
+            if revLicenseType() is "community" then
+               put revEnvironmentBinariesPath() & slash & "standalone-community" into tDeployInfo["engine"]
```
livecodefraser, line 117: Since the Business license has been added, all standalone engine files are named in title-case ("Standalone-Community" rather than "standalone-community"). The check should otherwise be okay, though - the business and indy standalone engines still use "commercial" as the suffix.

livecodeali, line 117: Ah ok, thanks. This is parked for now anyway, so I'll just revert to WIP.

PR #190, opened 2015-04-23
[[ Tutorial ]] LiveCode interactive tutorials

livecodeali:


livecodeali:
At this point, everything works but there are two implementation details which currently go against new IDE policy, namely
- some of the messages received by the tutorial stack are not received from the IDE library
- some of the highlighting (property inspector, tools) are not routed through the IDE library

livecodeali:
Still to be implemented in this branch 
- developer tutorial
- tweak flow for smaller screens

PR #202, opened 2015-05-06
[[ Bug 10722 ]] Port forward fix for bug 10722

livecodeali:
Previously when right-clicking a handler name to get a contextual menu in the script editor, "Go to definition" was disabled with gRevDevelopment not true. Removing this code makes "Go to definition" enabled all the time.


livecodeali:
The develop version of this: https://github.com/runrev/livecode-ide/pull/201

PR #214, opened 2015-05-20
[Player_colors] Locked: revpalette.rev

livecodepanos:
Added the option to set the backcolor of a player from the Property Inspector (the backcolor of a player is the color of the controller icons)


livecodepanos:
This pull request comes together with https://github.com/runrev/livecode/pull/2284

PR #217, opened 2015-05-21
[[Bugfix 15229]] - Message box stealing focus on open

runrevben:
Caused by the palette behaviour. The reason for this was the use of "start editing group". It was not possible to "create widget as xxx in group yyy". Ali has added this so IDE will break unless Ali's update is merged in.


livecodeali:
https://github.com/runrev/livecode/pull/2300 needs to be merged before this can go into develop.

PR #220, opened 2015-05-21
[[ bug 15290 ]] IDE frontscript not passing resumeStack

runrevben:


runrevben:
http://quality.runrev.com/show_bug.cgi?id=15290

PR #223, opened 2015-05-22
[[ Bug 14729 ]] Provide default icons for extension builder

livecodeali:


livecodeali:
Script of extension builder plugin changed as follows:

```
   # Load icons
   # AL-2015-05-22: [[ Bug 14729 ]] Display default widget and library icons if none are provided
   if there is a file pData["icon"] then 
      put URL ("binfile:" & pData["icon"]) into image "icon" of me
   else
      put  URL ("binfile:" & revIDEDefaultExtensionIcon(pData["type"], false)) into image "icon" of me
   end if

   if there is a file pData["retina_icon"] then 
      put URL ("binfile:" & pData["retina_icon"]) into image "icon_retina" of me
   else
      put URL ("binfile:" & revIDEDefaultExtensionIcon(pData["type"], true)) into image "icon_retina" of me
   end if
```

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -3874,13 +3874,21 @@ function revIDEExtensionFolders
...
+         end if
       case "widget"
       default
-         return revIDESpecialFolderPath("images") & slash & "extension-widget.png"
+         if pRetina is true then
+            return revIDESpecialFolderPath("images") & slash & "extension-widget@extra-high.png"
+         else
+            return revIDESpecialFolderPath("images") & slash & "extension-widget.png"
+         end if
    end switch
```
runrevben, line 3892: I'm not 100% sure about this. The engine auto selects if a platform is retina and seeks the appropriate @size.. Should we be checking for retina case here? would setting the image path to "extension-widget.png" suffice?

livecodeali, line 3892: Well, it's also there to provide the paths for the icon and retina icon when packaging the widget, so from that point of view it's correct.

In the case of the extension builder plugin it's taking the binary data from the file and settings the imagedata to it, so it won't be affected by the engine's retina stuff. We could set the image path instead, but isn't the idea to display the icons that have been selected as the icon and retina icon, and therefore display them as they actually are? Otherwise they would both be retina on retina machines, and both non-retina on non-retina. That's fine if that's what we want, I'm not quite sure.

PR #231, opened 2015-05-27
[[ Menubar UI ]] Add scriptified version of menubar UI

livecodeali:
This is an initial pass at replacing the menubar with a script-only stack.

Currently it just mimics the existing menubar UI and uses the menubar as its back end. Gradually we will move functionality over from the old menu bar.


livecodeali:
The gradual approach was found wanting on non-mac systems due to actual menubar behavior on those platforms, so once finished this branch will be a complete scriptification of the menubar.

PR #232, opened 2015-05-28
[[ Bug 14788 ]] Added lock screens in the frontscript backspaceKey and deleteKey handlers

livecodeali:
backported from this https://github.com/runrev/livecode-ide/pull/229


livecodesebastien:
@livecode-vulcan review ok e59e621

PR #235, opened 2015-06-03
Changes in home.rev and revStartCenter.livecode

livecodepanos:
CHANGES:

stack home.rev:
- Reverted previous changes. Now, Quicktime is loaded before revStartCenter windows appears (i.e as it was before)

stack revStartCenter:
- on preopenstack: Removed 'set the dontuseqt to true'

card createitNoQt:
- on preopencard: Removed 'set the dontuseqt to true'
- on setControls: Added 'set the dontuseqt of player "winPlayer" to true'


runrevmark:
Reviewed-by: @runrevmark

PR #237, opened 2015-06-05
[[ Menubar ]] Plumb in text menu

livecodeali:


runrevben:
I like the abstraction with all text styling running through a central function.

PR #238, opened 2015-06-05
[[ Menubar ]] Plumb in tools menu

livecodeali:


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -3327,15 +3342,22 @@ command revIDEOpenPalette pPaletteName
          break
       case "project browser"
       case "extension manager"
+      case "menu builder"
+      case "extension builder"
          modeless stack tStackName
          break
       default
```
runrevben, line 3349: Should "default" not move down to above the "return error"

livecodeali, line 3349: Yes, it doesn't make a difference functionally, but you're right, default as last item is better coding style.

livecodeali, line 3349: In fact, there is no point in the "search engine" case being there.

PR #244, opened 2015-06-08
Some final tweaks on the windows startcenter player

livecodepanos:
FIXES:
1.Previously, if the player was playing and the currentTime changed (either by clicking the
FF/FR buttons, or by clicking on the controller rect), then the player stopped. Now this is fixed, and the player preserved its state (playing or paused)

2.Previously the player did not respond when clicking on the controller rect


livecodepanos:
Ahh, I changed the user name and email settings only for the runrev/livecode repo, so it still has Michael's name since this commit is in runrev/livecode-ide. Sorry @runrevmichael I will fix it now :)

PR #245, opened 2015-06-09
[[ Menubar ]] Plumb in Object menu and contextual menus

livecodeali:


runrevben:
Cursory skim as changes are so wide ranging. Test procedure in place to go through code paths of all menu items following this merge.

PR #253, opened 2015-06-10
Sets the player name to empty, to let it hide (know bug)

livecodesebastien:
the issue only occurs on 7.0, but we want to keep the revstartcenter the same across the versions


livecodesebastien:
Only change:
`set the filename of player "winPlayer" to empty`
added before setting its visibility to false, in hideWinPlayer

livecodepanos:
@livecode-vulcan review ok a320478

PR #266, opened 2015-06-12
[[ Inspector ]] Ignore props in inspector when no editor specified

livecodeali:


livecodeali:
We've decided to deal with this differently - there is a property metadata property called user_visible which is true by default - widget devs should prevent properties being displayed by setting this to false in the widget metadata.

PR #269, opened 2015-06-15
[[ Inspector ]] Add mechanism to allow enum options to be specified by a property of the target object

livecodeali:


livecodeali:
For example, the navbar widget could specify property metadata:

metadata selectedLabel.editor is "com.livecode.pi.enum"
metadata selectedLabel.options is "property:navLabels"

so that the enum editor is populated with the correct data.

PR #299, opened 2015-06-18
[[ Menubar ]] Kill menubar flicker

livecodeali:
Fixed the menubar flicker by removing a "show me" from the updateMenubarPreference function. This function needs to hide the menubar on mac if preferences dictate that neither toolbar icons nor text should be shown.

Reinstated the 'show me' in the idePreferenceChanged handler, so that it is not called during preOpenStack.

Unfortunately, messages are not sent to hidden stacks, so had to add a workaround to the ide library to send messages to the menubar even if it is hidden.

I'm open to suggestions about a better way to go about that.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -1416,16 +1416,19 @@ on revIDEMessageSend pMessage, pEffectedTarget
          revIDEUnsubscribeAll tObjectID
          next repeat
       end if
-
-      # If the stack isn't visible on screen, don't send a message
+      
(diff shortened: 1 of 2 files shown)


livecodeali:
Potentially we should remove that whole clause from revIDEMessageSend, and make it individual palettes' / the ide library's responsibility to unsubscribe from messages when they are hidden.

runrevben:
Plan is to update the IDE message sending to always send. IDE palettes will be responsible for deregistering for messages. Merging this fix now to solve to presenting flicker but some changes will be made in a subsequent pull request.

PR #343, opened 2015-07-07
revIDESaveStack was not using the effective filename of the target st…

trevordevore:
…ack. This would cause the <Save As> dialog to appear every time you tried to save a script in a substack from the script editor.


livecodeali:
Good spot, thanks Trevor.

PR #344, opened 2015-07-07
[[ Bug ]] Update LCB error number

livecodeali:


livecodeali:
Do not merge until this is merged: https://github.com/runrev/livecode/pull/2543

runrevmark:
Could we compute this dynamically by searching through the execution error strings?

livecodeali:
Good idea!

PR #352, opened 2015-07-09
[[ MissingSampleMov ]] Use the appropriate case for Sample.mov in Ins…

livecodesebastien:
…taller/package.txt


peter-b:
Yay for making installers on case-sensitive filesystems!

@livecode-vulcan review ok f989c74

PR #378, opened 2015-07-17
[[ Bug ]] Fix data grid library after stricter repeat parsing merged

livecodeali:
After the stricter repeat parsing patch (https://github.com/runrev/livecode/pull/2428) was merged into develop, the data grid library no longer compiled. This fixes that issue, ported to 6.7 so as to keep ide stacks shared by develop and develop-6.7 in sync.

The only change here is to remove a `then` from the end of a `repeat` line.


livecodesebastien:
@livecode-vulcan review ok ea48d91

livecode-vulcan:
💙 review by @livecodesebastien ok ea48d91ea7cac703f6e0ac1f0e599fca075984af

PR #389, opened 2015-07-21
[[Cleanup]] Ignore generated documentation files.

peter-b:
Makes rebases on the runrev/livecode repository slightly less painful.


livecodeali:
These should be going away shortly once I can get the dictionary to work having them in a per-user application folder. But probably wise to gitignore them until that point.

PR #400, opened 2015-07-24
[[ Standalone Extensions ]] Add childWidget metadata for standalone inclusions

livecodeali:
We might want to implement something to automatically include child widgets in a standalone so that the user doesn't have to select them.


runrevben:
@livecode-vulcan review not ok 63433e4

PR #409, opened 2015-07-30
[[ Bug ]] Load extension with resource path if present

livecodeali:


runrevmark:
@livecode-vulcan review ok d0706e7

PR #411, opened 2015-07-30
[[ Bug ]] Include resource folder recursively when packaging extensions

livecodeali:


runrevmark:
@livecode-vulcan review ok cfdd76a

PR #412, opened 2015-07-30
[[ Docs ]] Make synonyms of current entry bold as well as homonyms

livecodeali:


peter-b:
Some discussion about whether `hasOwnProperty()` should be used or not, but that should probably be addressed separately. :ok_hand: 

@livecode-vulcan review ok 034b3a9

PR #415, opened 2015-07-31
[[ Docs ]] Rebuild distributed docs when building from source if there are docs changes

livecodeali:
Ideally updates to docs will be immediately reflected in the dictionary when building from source. Currently you have to open all the builder stacks and click build docs.

The script and builder dictionaries are cached separately so that the dictionary is not rebuilt every time an lcb file is changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,5 +66,5 @@ environment_log.txt
 
 # Documentation #
 #################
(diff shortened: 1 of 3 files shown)


peter-b:
@livecodeali Reviewed. Thanks!

PR #421, opened 2015-08-13
revdebuggerlibrary: More aggressively filter problematic var names

peter-b:
On Linux systems, environment variables are just nul-terminated
strings, and only by convention have the format `name=value`.  This
means that variable names obtained from the environment (i.e. global
`$name` variables) may contain special characters, especially `(` and
`)`.  When environment variable names with special characters are
encountered, it causes subtle debugger breakage which usually
manifests itself as "the message box doesn't work".

@runrevmark originally suggested fixing this problem by filtering out
_all_ variables beginning with `$` at this step.  However, that seemed
like a bit of a blunt instrument, since it may actually be useful to
examine environment variables that might be affecting the behaviour of
the program -- and it's hard to do that in the absence of a `the
environmentVariables` syntax or equivalent (analogous to the
recently-added `the commandArguments` syntax).

### DIFF
--- a/Toolset/libraries/revdebuggerlibrary.livecodescript
+++ b/Toolset/libraries/revdebuggerlibrary.livecodescript
@@ -237,16 +237,22 @@ function revDebuggerGlobalNames
 end revDebuggerGlobalNames
 
 # Returns
-#   A list of the globals with valid names. I.e not the ones with (x86) in them on Vista 64bit.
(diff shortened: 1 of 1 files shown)


peter-b:
Okay -- on further discussion with @runrevmark  this should be fixed in the engine by not generating environment variable `$<name>` globals unless `<name>` is a valid token.

PR #422, opened 2015-08-14
[env-vars] Make revDebuggerValidGlobalNames into a no-op.

peter-b:
Since runrev/livecode#2716 has been merged, the constraints on naming
of `$<name>` global variables are much tighter.  The filtering that
`revDebuggerValidGlobalNames` does is no longer required.

However, because the function is used elsewhere (in particular, in the
script editor's variables inspector pane), it can't be removed
entirely.


runrevmark:
Yay! Code deletion :)

peter-b:
@runrevmark I just wish I hadn't had to add several times as much code in order to enable the deletion!

runrevmark:
@peter-b: At least the code is in the right place now though :)

PR #423, opened 2015-08-17
[[ Bug 15522 ]] Apply fix by Bernd Niggemann for broken cell offset w…

livecodesebastien:
…hen field loc is not 0,0


livecodesebastien:
The pull request should be made against develop-6.7

PR #450, opened 2015-08-28
[[ Bug 15804 ]] Add missing break to fix Reshape Graphic option from Object menu

BerndN:
Bug 15804 - FIX: selecting "Reshape Graphic" from Object menu causes a new card to be added

missing "break" causes new card to be created.

Bernd Niggemann


livecodeali:
Thanks Bernd!

I have edited the title of the pull request to match our 'style'. This is just so that we can tell what the content of the pull request is from the title.

PR #482, opened 2015-09-07
[[ Bug 15884 ]] Reinstate ability to view and edit variables in a popup stack

livecodeali:
Double-clicking on a row, or clicking on the 'open in external window' icon in the variables tree view now launches the visualization stack as it did previously.

We can't currently edit inline so double-clicking launches the stack even for short and single line variable values.

Also moved the variables pane behavior to a script-only stack.


runrevben:
@livecode-vulcan review ok 48d6c7c

livecode-vulcan:
💙 review by @runrevben ok 48d6c7ca58cfc55249fa5db32d417193ec3744e6

PR #485, opened 2015-09-09
Merge develop 7.0 09.09.16

livecodeali:
Conflicts almost exclusively in code that doesn't exist in 8.0 anymore.

Removed some logging from IDE library that used to be in menu bar code.


livecodeali:
Ah, this appears to have my svg icon property editor in it :-1:

PR #488, opened 2015-09-09
[[ Bug 15745 ]] Ensure PI doesn't appear off screen

livecodeali:


runrevben:
This was a feature that we disabled. I think there was a revIDEEnsureOnscreen function that checked where the use last put it as well as whether it was on screen. I think we wanted to refine the palettes before working on the placement. Is this fix redundant if we reenable that feature?

livecodeali:
Sort of, although this one tries to place the PI to the left of the stack if it can't put it at the right, so slightly different

livecodeali:
That's what revInternal__ConstrainStack does, doesn't it? Ensure on screen?

runrevben:
Yeah I think so. The revIDEEnsureOnscreen looks at prefs and stores positions and sizes on close. I think it needed refinement. If you're happy with the fix as it is, I can review ok.

livecodeali:
I don't see a revIDEEnsureOnscreen, but it's probably a good idea to ensure arbitrary palettes are onscreen after checking the prefs.

runrevben:
@livecode-vulcan review ok 4b826d2

livecode-vulcan:
💙 review by @runrevben ok 4b826d23572f1e2d8868860032c3a78d9121692a

PR #491, opened 2015-09-14
[[ Bug 15926 ]] Add filter field and show/hide global/environment variable options to the script editor variables pane

techstrategies:
Bug report submitted: http://quality.runrev.com/show_bug.cgi?id=15926


livecodeali:
Added the objects here:

https://github.com/livecode/livecode-ide/pull/492

PR #495, opened 2015-09-16
faster locateHandler

mwieder:
A faster and more modern locateHandler, taking advantage of the new regex filter syntax. Also internalized the switch statement since it's only called from locateHandler. I think it makes the whole thing easier to read and understand as well.

The pProposedPosition parameter isn't needed, so I think removing it from the two places in other scripts where locateHandler is called would also be an improvement.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
@@ -952,35 +952,51 @@ command locateHandler pName, pProposedPosition, pType
    local tScript
    local tFilteredScript
    local tType
+   local tMatchRegex
    -- this allows us to ignore the 'private' modifier
    constant kRegex = "^(?:private\s)?"
   
    put the text of field "Script" of me into tScript
-   switch char -1 of pType -- ignore a "P" modifier if present
-   	case "M"
-   		put "(on|command)" into tType
-   		break
-   	case "F"
-   		put "function" into tType
-   		break
-   	case "G"
-   		put "getprop" into tType
-   		break
-   	case "S"
-   		put "setprop" into tType
-   		break
-   	case "B"
-   		put "before" into tType
-   		break
-   	case "A"
-   		put "after" into tType
-   		break
-   end switch
-   filter tScript with regex (kRegex & tType & "\s" & pName & "\b.*" ) into tFilteredScript
-   put lineoffset(tFilteredScript, tScript) into tResult
-   if tResult is 0 then
-   	put empty into tResult
-   end if
+	switch char -1 of pType -- ignore a "P" modifier if present
+		case "M"
+				put "(on|command)" into tType
+				break
+		case "F"
+				put "function" into tType
+				break
+		case "G"
+				put "getprop" into tType
+				break
+		case "S"
+				put "setprop" into tType
+				break
+		case "B"
+				put "before" into tType
+				break
+		case "A"
+				put "after" into tType
+				break
+	end switch
+   put kRegex & tType & "\s" & pName & "\b.*" into tMatchRegex
+   --if matchText(line pProposedPosition of tScript, tMatchRegex) then
+   	-- if the script is already compiled we will get a match
+  		--put pProposedPosition into tResult
+   --else
+   	-- otherwise we ignore pProposedPosition and find the real line number
+		-- filter down to just the line we're looking for
+		filter tScript with regex tMatchRegex into tFilteredScript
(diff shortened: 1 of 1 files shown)


livecodeali:
Doesn't this make locateHandler slower in the (potentially) more common case that the script is compiled (and thus pProposedPosition is correct)? 

I think the best thing would be for it to check somehow if the current script was compiled, and if so just return pProposedLine, and if not use your updated code.

I don't know if there exists anything in the script editor currently to query if the current open tab has been compiled aside from checking if the apply button is enabled. Perhaps the script tabs behavior should keep track of it?

livecodeali:
If pProposedPosition is not reliable then maybe something like

```
 local tMatchRegex
 put kRegex & tType & "\s" & pName & "\b.*" into tMatchRegex
 if matchText(line pProposedPosition of tScript, tMatchRegex) then
    return pProposedPosition
 end if
```

would make sense.

mwieder:
Here's a somewhat revised version. I'm not convinced this makes things slower in the more probable case, since 'line x of text' and 'lineoffset(x, text)' are relatively similar functions, and the filter function is blazingly fast while matchtext is a bit less so. I did replace my call to lineoffset with a repeat loop, based on the benchmarks we did last month in a web forum.

That said, I tried several different scenarios of compiled/non-compiled/edited/saved versions of a test script, and the pProposedPosition value was correct in all cases. Is it possible that the situation that Ollie originally crafted this function for in 2009 is no longer a problem?

livecodeali:
Ok I ran some tests.

The first version of locateHandler I used was your original PR.
The second version is your revised PR
The third version uses the commented out part of your revised PR
The fourth version does the same as the third version, but with

```
   ...
   local tProposedLine
   put line pProposedPosition of pScript into tProposedLine
   filter tProposedLine with regex tMatchRegex into tFilteredScript

   if tFIlteredScript is not empty then
      --if the script is already compiled we will get a match
      put pProposedPosition into tResult
  else
```

instead of using matchText.

I ran locateHandler 1000 times each with different parameters.
Here are the results:

```
Test 1: Long script, handler late on in script, correct line passed
revIDEGenerateDistributedGuide line 8111 of script of revIDELibrary

1 millisecs : 9980
2 millisecs : 15580
3 millisecs : 1518
4 millisecs : 1526

Test 2: Long script, handler late on in script, incorrect line passed
revIDEGenerateDistributedGuide line 8110
1 millisecs : 10153
2 millisecs : 15852
3 millisecs : 17388
4 millisecs : 17395

Test 3: Long script, handler early on in script, correct line passed
__resolvePopupData line 254 (correct)
1 millisecs : 7082
2 millisecs : 7048
3 millisecs : 62
4 millisecs : 64

Test 4: Long script, handler early on in script, incorrect line passed
__resolvePopupData line 253 (incorrect)
1 millisecs : 6981
2 millisecs : 7057
3 millisecs : 7079
4 millisecs : 7034

Test 5: Short script, handler in middle of script, correct line passed
ideSelectedObjectChanged line 346 (correct)
1 millisecs : 462
2 millisecs : 717
3 millisecs : 78
4 millisecs : 78

Test 6: Short script, handler in middle of script, incorrect line passed
ideSelectedObjectChanged line 345 (incorrect)
1 millisecs : 465
2 millisecs : 720
3 millisecs : 784
4 millisecs : 791
```

So the matchtext version is the fastest if the proposed line is correct, and your first version is the fastest if the proposed line is incorrect. Of course, this is all moot if we can work out how the pProposedPosition parameter is being determined, and verify that is is always correct anyway.

mwieder:
I'd be quite happy to find that the proposed line is now correct in all cases, and thus eliminate having to call this routine at all. I found that was true in my testing, and was unable to come up with a scenario in which the proposed line was wrong. Maybe Mark Waddingham has some historical data to add as to why this was necessary in the first place.

I do find it interesting that your benchmarks show a complete turnaround from what we discovered in the community process on the web forum. Instructive indeed.

livecodeali:
Unfortunately it isn't - I just tried compiling a script, and then adding a bunch of lines at the top (with locateHandler just returning pProposedPosition), and it selects the wrong line.

What is the community process on the web forum? The reason my suggested version is faster isn't so much a result of superior individual operations, it's just that fetching a particular line and checking if it matches regex is less work that filtering all the lines with regex and finding the offset of the line found.

mwieder:
I'll dig up a reference to the forum conversation for you for offline reading.

But if the edge case still exists I think I'd like to rescind this pull request and just leave things as they are. I'm mostly just playing around with the code here, and if I don't have anything significant to contribute to this, I'd rather not waste your time any more.

PR #511, opened 2015-09-23
[[ Bug ]] Remove existing hack for builtin modules and replace with hard-coded list

livecodeali:


runrevben:
Would it be possible to do this as metadata in the extension itself? The would save us having to update this list?

livecodeali:
Unfortunately not - the built in modules don't have manifests or anything. I think there should be a better solution to this, but for now replacing a silly hack with a sensible hack seems like the best option.

runrevben:
OK, it's certainly better with the change you've made.

runrevben:
@livecode-vulcan review ok 063b9b4

livecode-vulcan:
💙 review by @runrevben ok 063b9b4e1c86513679a2aaf28835f93bafd2c649

PR #515, opened 2015-09-24
Bugfix refactor ide stack check

livecodeali:
- Fixes bug 16010
- Uses IDE stack check function in a number of places in the IDE
- Improves performance of IDE stack check function
- Removes a bunch of places in the IDE message handler library which were fetching the long id of the target twice for no reason


runrevben:
This is a good change! 
@livecode-vulcan review ok 3b1d3c2

livecode-vulcan:
💙 review by @runrevben ok 3b1d3c27e887690fe26dd22de41c3b3f90891dac

PR #516, opened 2015-09-25
[[ Bug 16013 ]] Clean up tools palette control drag

livecodeali:
- Removed some inefficient code
- Ensure existing bg color of target stack is restored
- Unhilite stacks when the mouse is no longer within the rect
- Set the loc of the revDragControl stack while it is invisible to prevent control appearing at top left of screen

### DIFF
--- a/Toolset/palettes/tools/revtools.livecodescript
+++ b/Toolset/palettes/tools/revtools.livecodescript
@@ -573,12 +573,12 @@ on mouseDown
    unlock screen
    
    # AL-2015-03-26: [[ Bug 14892 ]] Simply set the visible of the revDragControl stack to true
+   set the loc of this stack to the screenmouseloc
    show this stack
    
    # Let the user drag around
-   local tExit, tCount, tValidDropStacks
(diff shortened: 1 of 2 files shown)


peter-b:
@livecode-vulcan review ok 09b9596

livecode-vulcan:
💙 review by @peter-b ok 09b95966a0de342d72abbb14f0acbcfda35f726a

peter-b:
@livecodeali Before merging this, could you please double-check that this makes a difference in release mode (so that we can be sure that the bug is fixed!)?

livecodeali:
Sorry, I should have clarified - the actual 'jerkiness' bug is caused by the tCount counter which only updates the loc of the revDragControl stack every 12 ticks of the repeat loop. So if you move the mouse very quickly there is a lag/jerk.

I can separate this out into two pull requests if that's necessary.

peter-b:
@livecodeali Oh, right. That makes sense. Sorry for missing that in the first place!

PR #519, opened 2015-09-25
[[ Bug 16025 ]] Use actual dg object rather than snapshot when dragging out

livecodeali:


runrevben:
Hey Ali, what was the thought behind this? Looks like a solid update which I can mark review ok.. just wondering what prompted the change.

livecodeali:
For some reason taking a snapshot of the datagrid doesn't work - so when datagrids are dragged out they are just a line. I can't work out why this is, so it seemed reasonable just to revert to the old behavior in that specific case.

runrevben:
Ah, makes complete sense!

runrevben:
@livecode-vulcan review ok aac5d7e

livecode-vulcan:
💙 review by @runrevben ok aac5d7ee18fc73069f06cbc66cb65df4210642f0

livecodeali:
There was a trivial merge conflict here.
@livecode-vulcan review by @runrevben ok 6303b53

livecode-vulcan:
💙 review by @runrevben ok 6303b535d8d10a8b17eda478f26873fdc05c3353

PR #521, opened 2015-09-26
[[ Bug 16000 ]] Don't allow lingering revNewButton message after IDE startup

livecodeali:
Since the revTools stack is script-only and creates buttons, it was triggering the newButton message, which in turn was queuing 'revNewButton' in the pendingmessages.

The long id of the target includes the full path to the IDE tools stack.

Therefore when a plugin (eg PowerTools) tries to replace the tools palette, this lingering message causes the IDE tools palette to be loaded into memory again, causing a naming conflict.

### DIFF
--- a/Toolset/libraries/revbackscriptlibrary.livecodescript
+++ b/Toolset/libraries/revbackscriptlibrary.livecodescript
@@ -689,7 +689,7 @@ end newGroup
 
 
 on newButton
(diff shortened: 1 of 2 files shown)


peter-b:
@livecode-vulcan review ok c4e4069

livecode-vulcan:
💙 review by @peter-b ok c4e406950cb904879b3f50d31db549abea116ef2

PR #537, opened 2015-10-01
[[ Bug 16065 ]] Remove divider in windows menubar, and ensure menubar prop of stack revMenuBar is set

livecodeali:


livecodeali:
The divider needs to be there for Linux, so this needs a slight rework.

PR #539, opened 2015-10-05
[[ Bug 16079 ]] Activate breakpoints before preOpenStack handlers are called

techstrategies:


livecodefraser:
Thanks!

@livecode-vulcan review okay 4f3b670

livecode-vulcan:
💙 review by @livecodefraser ok 4f3b67088c25e85a6b96c530c66db44683b4e770

livecodefraser:
@runrevmark is there a reason this was done in openStack previously?

livecodesebastien:
@techstrategies: In the next times you propose a fix (thanks again for your help!), can you simply delete the code that you remove, instead of commenting it? That leaves the source a bit cleaner, especially if the previous code was wrong :) Thank you for the patch anyway!

techstrategies:
@livecodesebastien no worries, will do!

PR #543, opened 2015-10-05
[[ Bug 16111 ]] RevSaveAs: build the path using the stack's filename,…

livecodesebastien:
… not it's stack name

### DIFF
--- a/Toolset/libraries/revbackscriptlibrary.livecodescript
+++ b/Toolset/libraries/revbackscriptlibrary.livecodescript
@@ -964,7 +964,8 @@ on revSaveAs pShortName, pSubstackOnly
       end if
    end if
    
-   put revFixPath(pShortName) into tStackName
+   // revFixPath expects to get the actual filename, not the stack's short name
+   put revFixPath(the filename of stack pShortName) into tStackName
    
    // MM-2012-03-09: Added new 5.5 stack file format to drop down.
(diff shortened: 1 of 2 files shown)


livecodeali:
I think this will work only if the stack already has a filename - otherwise you just get ".livecode" back as the result of revFixPath and it still opens the file dialog in an IDE folder.

livecodeali:
I think we need to find where the defaultFolder has been set without being reset too.
@livecode-vulcan review not okay eccc775

livecode-vulcan:
💔 review by @livecodeali not ok eccc77567a1f3b108c46089af09bdba88ce0863b

livecodepanos:
I think in LC 6.7.x /7.x the defaultFolder is set on `startup` handler of the `Home` stack. It defaults to `/Applications` on OSX. In LC 8, these lines are missing from the `startup` handler. However, I tried to add them but the result is the same as described in the bug report. So my guess is that the defaultFolder is being set somewhere else too.

peter-b:
@livecodepanos It looks to me as if it might be being set while loading LCB extensions and not restored as expected.

livecodesebastien:
After a better look, the issue does not come at all from the IDE, but from the engine.
Superseded by https://github.com/livecode/livecode/pull/3024

PR #547, opened 2015-10-05
[[ Bug 14502 ]] Ensure answer dialog is dismissed with correct data when return pressed on Linux

livecodeali:
Fix implemented as suggested by Richard Gaskin to trigger click at the loc of relevant button on return key, namely to replace

```
on returnKey
  pass returnKey to top
end returnKey
```

with

```
on returnKey
  click at the loc of button id 1240
end returnKey
```

`button id 1240` is the default button of the answer dialog stack, so this reference is always correct.


peter-b:
@livecodeali Is it possible to reference the button by name, or is that unreliable in this context? I have literally no idea whether that's a sane question to ask...

livecodeali:
@peter-b No - the name is changed depending on the answer command. If you execute

`answer "Hello" with "Hello" and "Goodbye"`, the default button will be called "Hello".

peter-b:
@livecodeali Oh, I see. I thought just the button captions were changed.

peter-b:
@livecode-vulcan review ok e893958

runrevmark:
@livecodeali: If a button is 'default' then pressing the return key should cause it to be triggered... So this could be an engine bug, or something in the answer dialog is stopping the default behavior from working correctly.

livecode-vulcan:
💙 review by @peter-b ok e8939582b2ceb93becff824c8dd429f66fcceaf2

runrevmark:
@livecode-vulcan review not ok e893958

livecode-vulcan:
💔 review by @runrevmark not ok e8939582b2ceb93becff824c8dd429f66fcceaf2

PR #551, opened 2015-10-06
[[ Bug 16140 ]] Make the variables pane default to not showing environment variables

techstrategies:


peter-b:
Why not add a new script editor preference to control whether environment variables are shown?  I.e. so there are separate "show global variables" and "show environment variables" preferences?

techstrategies:
@peter-b My understanding (and I could be wrong) from Ali was that with the revpreferencesgui stack currently a binary stack and identical to the one in LC7, that it was preferable not to have to fork it from the version 7 one.

livecodeali:
I think Peter is talking about a script editor preference without a corresponding GUI element, which just remembers your previous setting (i.e. something like)

```
 on setEnvDisplay pShowEnvGlobals
    set the cShowEnvGlobals of me to pShowEnvGlobals
    sePrefSet "variables,showenv", pShowEnvGlobals
    send "updateVariableDisplay" to me in 0 millisecs
 end setEnvDisplay
```

and then set the hilite accordingly on load.

mwieder:
Am I missing something here? There is already a checkbox 'Show Environment Vars' in dp-6 which is unchecked by default. So the variables pane already defaults to not showing environment variables.

livecodeali:
@mwieder I think the issue is that if you have you preference set to showing globals, it also show env vars

mwieder:
Ah. Thanks - yes, that makes sense then.

techstrategies:
Ah - yeah, that sounds like a good idea.  Sorry for not catching on quicker!  PR updated accordingly.

livecodeali:
Perfect, thanks :+1: 
@livecode-vulcan review ok 2925496

livecode-vulcan:
💙 review by @livecodeali ok 2925496efe81bcbbe4fc666dc33717afde05b67b

PR #558, opened 2015-10-06
[[ Bug 14502 ]] Don't set Answer Dialog OK button's traversalOn to false on Linux

livecodeali:
This causes it not to be the curcard's kfocused control, and hence prevents
return from triggering it's mouseUp

Changes:
In the platform switch in answer dialog's preopenstack handler, just removed the line
`set the traversalOn of button i to false`


peter-b:
@livecode-vulcan review ok c753c1f

livecode-vulcan:
💙 review by @peter-b ok c753c1f02a321d9f4362754fbd4a0a676adf6823

PR #565, opened 2015-10-14
[Bug 15887] Handle selectionChanged sent by the engine to the scriptEditor

livecodepanos:
This fixes a regression caused by the fix to bug 7217. That fix was planned to be included in LC 6.7.8 RC1, but for some reason it is already included in LC 7.1.0 RC1. According to that fix, the engine should send a "selectionChanged" message to the field when navigating using arrow keys, regardless of whether the Shift key is held down. That resulted in having four different `selectionChanged` messages:

1) A `selectionChanged`  that is sent to the field by the engine, and has no parameters. On LC 7.1.0 DP1 (i.e before the fix to bug 7217), this was sent to the field ONLY when navigating using arrow keys AND the Shift key was down. The handler `selectionChanged pArrowKey` of button "Template Editor” handled this case with:

```
if the shiftKey is "down" then
   exit selectionChanged
end if
```

2) A `selectionChanged pArrowKey` that is sent by the `rawKeyUp pKey` handler:

`send "selectionChanged tKeyName" to me in 0 milliseconds`

3) A `selectionChanged` (with no parameters) that is sent by the `rawKeyDown pKey` handler, to handle the Home and End keys:

`send "selectionChanged" to me in 0 milliseconds`

The handler `selectionChanged pArrowKey` of button "Template Editor" handled this case with:

```
switch pArrowKey
      case "left"
         …
      case "right"
         …
      case "up"
      case "down"
         …
      case "empty"
      default
         caretUpdate tField, tScript
         break
   end switch
```

All these 3 cases were handled by the `selectionChanged pArrowKey`  handler. Now, after the fix to bug 7217, there is another `selectionChanged` message (with no parameters) that is sent by the engine. Call this case (4). Currently, case (4) is handled as case (3), since pArrowKey==empty, but it should be handled as case (1).

So this patch addresses this problem by:
1. Renaming the handler `selectionChanged pArrowKey` to `selectionChangedByArrowKey pArrowKey` so as to keep its existing functionality to handle cases (2) and (3), 
2. adding another `selectionChanged` handler that will handle cases (1) and (4).

The new handler `selectionChanged` just updates the `sLastNonEmptySelection` and `sLastSelectedChunk`, just the way it was done in the old `selectionChanged pArrowKey` handler (now renamed to `selectionChangedByArrowKey pArrowKey`).


livecodepanos:
Binary diff:
https://gist.github.com/livecodepanos/028f80db16ab28539a0c

livecodepanos:
I am closing this PR, and I will submit a new one, since this patch includes changes in a binary file, which has already been updated since this PR

PR #577, opened 2015-10-22
[[ Bug 16176 ]] Make inspector respect language names preference

livecodeali:
At the moment we can't make this update 'live' without either a
hack, or changing the binary preferences gui stack - I'd rather not
do the latter until we are sure there will be no changes coming up
from 6.x/7.x

Similarly, the custom property name needs to change as `cLanguageNames`
doesn't really cut it for me as a description of the property. Also
for some reason it is a custom property set on _card_ 1 of stack
"revPreferences" instead of just the stack.

Finally, I don't see any reason for an associated and equally
non-descriptive global `gRevLanguageNames`, except because the custom
property can't be accessed by the normal API as it's set on the card.

In short, there's a lot of ways to improve this when we can.

### DIFF
--- a/Toolset/palettes/behaviors/revinspectorbehavior.livecodescript
+++ b/Toolset/palettes/behaviors/revinspectorbehavior.livecodescript
@@ -335,9 +335,6 @@ on inspectorGenerateGroups pGroupList
       local tName
       put the short name of group tElement["label"] of me into tName
       if exists(group tElement["label"] of me) then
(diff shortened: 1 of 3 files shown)


peter-b:
Nevertheless, this seems like it solves a problem for the time being, even if it needs to be revisited in the future!

@livecode-vulcan review ok b328868

livecode-vulcan:
💙 review by @peter-b ok b32886892f1520bbe2017073db2fd38ccd14300e

PR #584, opened 2015-10-23
[[ Bug 16176 ]] Fix some issues caused by interaction between PI fixes

livecodeali:
Namely, the fix for bug 16176 delayed the setting of the rowLabel until after
the property was registered with the editor. This fixes that by assigning the
editor label properly. Also tidied up the showing of the rowLabel, hiding the
field instead of emptying it.


peter-b:
@livecode-vulcan review ok c815466

livecode-vulcan:
💙 review by @peter-b ok c815466576b0cf9042472c3c117504ca284860ac

PR #585, opened 2015-10-23
[[ iOS9.1 ]] Update Installer/package.txt to add iOS 9.1 engines

livecodesebastien:


livecodesebastien:
Should be against `release-6.7.8`

PR #594, opened 2015-10-23
Modified the Application Browser to run as a plugin.

runrevelanor:
The revApplicationOverview stack was referenced directly in a number of places so the backscript library, frontscript library and preferences.gui had to be updated to check the stack existed.


livecodeali:
Hi Elanor,
Instead of doing things this way round, it would be better to make the Application Browser subscribe to the new ide versions of these messages (ideNameChanged, ideCardDeleted etc) and then all the explicit references to it can be completely removed - that's how we're going to want third-party plugins to be developed.

I think we need to rework the ide<message> messages a bit anyway, as we need them to have the same parameters as the original and sometimes more (eg ideNewCard should maybe have the name of the card as a parameter).

livecodeali:
As discussed offline, we're putting this in the plugins folder as is and then will think about updating to the new message system before RC 1
@livecode-vulcan review ok a31d6af

livecode-vulcan:
💙 review by @livecodeali ok a31d6afc717a45e616a67d17a3bd81758c0c828d

livecodeali:
@livecode-vulcan review ok 8fc8d2b

livecode-vulcan:
💙 review by @livecodeali ok 8fc8d2bdf3a5cf53df9d1c7fca80db83f85d2ec8

PR #610, opened 2015-10-28
[[ Bug 16281 ]] If msg box string looks like a command, try to execute it

livecodeali:
Engine commands should have taken precedence over other interpretations of
the message box string.

We _really_ need to work out using library stacks in the lcs test system
and write tests to ensure the message box does what is expected.


peter-b:
@livecode-vulcan review ok fdb7373

livecode-vulcan:
💙 review by @peter-b ok fdb737327e5debe06c4be2769541d9796b8a4430

PR #615, opened 2015-10-30
[Tests] Make messagebox TestSetup handler use revised API

peter-b:
The test library was changed since commit 6e1f2eff1c99 so that the
path-related helper functions now also start with "Test".


livecodeali:
@livecode-vulcan review ok eeac679

livecode-vulcan:
💙 review by @livecodeali ok eeac6791a080231541f604b730acd3eb0c6cbe24

PR #617, opened 2015-10-30
[[ Message Box ]] Add tests for behavior when executing code in single line message box

livecodeali:
Since the single line message box does various bits of processing, autocompletion and context switching to execute the code that is entered, it requires tests to ensure these execute as expected in various situations.

In order to facilitate this, the core code that does all the intelligent bits has been refactored into the ide library.


peter-b:
@livecode-vulcan review ok 0f864a5

livecode-vulcan:
💙 review by @peter-b ok 0f864a5ded2776a839be73da287ec7ed2e763dbc

PR #618, opened 2015-10-30
[Bug 16304] Make sure screen is not locked when deleting text using visual effects

livecodepanos:
The `backspaceKey` and `deleteKey` handlers in revfrontscriptlibrary had an unbalanced `lock screen` command. This resulted in the screen being locked when deleting text using visual effects, thus the visual effect was not visible.


runrevmark:
@livecode-vulcan review ok 1f52f75

livecode-vulcan:
💙 review by @runrevmark ok 1f52f7527f5fb40159656f0c72c92452f63dfa34

PR #620, opened 2015-10-31
[[ IDE Library ]] Add some docs and tests for IDE library API

livecodeali:


peter-b:
Random question: am I write in thinking that `/** */` will be interpreted as a doc comment (i.e. I don't need two asterisks at the end as well as at the start)?

peter-b:
Second random question: should we have some style guidelines for writing API documentation, at the level of, "The summary should be a brief sentence written in the imperative voice"?

peter-b:
@livecode-vulcan review ok efd5365

livecode-vulcan:
💙 review by @peter-b ok efd5365b9482db4aef41f6ad9c81f8aa8be9c7fc

livecodeali:
To the first random question, yes you are correct. I just think it's nicer to have them balanced, and so that you can see that the end is supposed to be the end of a docs comment.

To the second, definitely!

PR #622, opened 2015-11-02
[[ Bug 16299 ]] Added checkbox to disable ATS on iOS S/B

livecodesebastien:
Xcode 7.0-introduced ATS prevents users to load URL that do not use HTTPS protocol.
That can be circumvented by adding an entry in the Plist file of the app - which can be asked for in the
iOS Standalone Settings (the default behaviour being to comply to Apple strict rules)


livecodesebastien:
Should be done against `release-6.7.8`

PR #636, opened 2015-11-06
[Bug 16371] Standalone settings cannot be changed - Locked:revstandalonesettings.8. 

livecodepanos:
When you are on OS X card of Standalone Settings, and you modify an existing fld (i.e. "identifier"), then the closeField/returnInField/tabKey of group OSX is called. All these handlers used to call the propertyChanged handler of group OSX. 

In LC 8, the propertyChanged handler of revFrontscriptLibrary was called, instead of the  propertyChanged handler of group OSX.

The same happens for card Windows (see the script of group Windows).

This fix includes renaming the "propertyChanged" handler of group OSX to "OSXpropertyChanged", and make sure to rename all the calls to it in the script of OSX group. (Similarly for Windows).


livecodepanos:
Binary diff:
https://gist.github.com/livecodepanos/755c67111dcb65f44a96

livecodepanos:
Binary Diff:

https://gist.github.com/livecodepanos/a61256cd2d66ebf635e9

livecodeali:
@livecode-vulcan review ok a9ccb17

livecode-vulcan:
💙 review by @livecodeali ok a9ccb17cebae325dff98272979ce9de6e717eab9

PR #645, opened 2015-11-10
[[ PubSub ]] Allow all messages to be per-object

livecodeali:
This tweaks the current publish and subscribe functionality so that a 'source object' can always be specified on subscription, so that messages from certain sources can be ignored. It also adds the ability for messages to have arbitrary numbers of parameters.


peter-b:
@livecodeali This seems to have acquired conflicts!

livecodeali:
I was thinking about using the params instead of the indexed array here, but it turns out that's not very good when the parameters passed have quotes and multiple lines :-(

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1601,33 +1658,58 @@ on revIDEMessageSend pMessage, pEffectedTarget
...
+      # Ensure any error doesn't prevent subsequent subscribers from receiving messages
+      try
+         do tDoString
+      catch tError
+         if tErrors is empty then
+            put tError into tErrors
+         else
+            put return & tError after tErrors
+         end if
+      end try
```
peter-b, line 1689: I dearly wish there was a way to `dispatch` with an indexed array of parameters, rather than having to construct some text to `do`.

peter-b:
@livecode-vulcan review ok 83e4ad9

livecode-vulcan:
💙 review by @peter-b ok 83e4ad9a3fa5c6e2703bdcda371d5bf79a7a050a

Toolset/libraries/revidelibrary.8.livecodescript
livecodeali, line 1689: Yeah - If `the params` escaped quotes and return characters then it might be possible to do something with that (i.e. `dispatch <message> with word 3 to -1 of the params` or something like that)

PR #649, opened 2015-11-11
[[ Bug 16398 ]] Fix three issues with custom property setting in PI

livecodeali:
- Ensure the current custompropertyset of the object is restored after setting
- Set the custom properties using the customproperties property - this
  prevents errors when a custom prop name in a particular set matches an engine
  property
- Allow return characters in the custom prop editor value field


peter-b:
@livecode-vulcan review ok 5bea81e

livecode-vulcan:
💙 review by @peter-b ok 5bea81eb6eec5210b1bfbd9fc1def6e59a353db8

PR #654, opened 2015-11-12
[[ Bug 15974 ]] Make 'effective' value text darker and italic

livecodeali:


peter-b:
@livecodeali It would be nice to have some sort of generic API for PI editors to fetch the "right" styling for things like effective values.

@livecode-vulcan review ok 3f39dc2

livecode-vulcan:
💙 review by @peter-b ok 3f39dc2d297a2f0a62034176b62dea9481188d42

livecodeali:
Indeed it would - I guess my thinking is that once these are replaced by field widgets then this sort of stuff will be done through their theming library anyway. Admittedly that's a pretty distant 'once'...

PR #659, opened 2015-11-16
[[ Bug 16423 ]] Make sure message box opens when 'put' command is executed

livecodeali:
The msgChanged handler was in the behavior which meant that it wasn't triggered until
the message box behavior was loaded (which happpens on preOpenStack). Moreover, msgChanged
did not `go stack the short name of me`, which it needs to if the stack is closed.

Added to revMessageBox.8.rev:

```
   if the mode of me is 0 then
      go stack (the short name of me)
   end if
```


livecodeali:
BTW I tried to write tests for this but the message box stack proved extremely difficult to write tests for due to its dependencies.

peter-b:
@livecode-vulcan review ok e0bbc39

livecode-vulcan:
💙 review by @peter-b ok e0bbc3925e822bdfdbf490c4f2a30847afbf3c93

PR #660, opened 2015-11-16
[[ Bug 16429 ]] Update custom props code to return 'per-object' array values

livecodeali:
The code which fetched custom properties for the property inspector hadn't
been updated to take into account the fact that it should be returned per-obejct
in order to allow for editor value conflicts.


peter-b:
It even has tests!!

@livecode-vulcan review ok d40c0e4

livecode-vulcan:
💙 review by @peter-b ok d40c0e4ece551f2dcbecf73ec634fa478e8cb5a3

PR #661, opened 2015-11-16
[[ Bug 16424 ]] Ensure geometry manager info isn't empty

livecodeali:
The property info array for the Geometry Manager section already
contained a Geometry Manager key with an empty array, causing the
union not to add the necessary info.


livecodeali:
I feel like this should work as it was, and there may be a bug with recursive array union...

livecodeali:
Ok, there are indeed issues with array union (and intersect):
See https://github.com/livecode/livecode/pull/3277

This PR suffices as a workaround in case the above doesn't make it into 8.0 DP 10.

peter-b:
Closing this because livecode/livecode#3277 fixes it differently.

PR #663, opened 2015-11-16
[[ Bug 16433 ]] Uninstall docs after updating the extension cache

livecodeali:
The extension docs use the cached install paths to find the documentation
source for installed extensions. The cache must therefore be updated before
regenerating the docs, otherwise the docs generator will try to set a default
folder that doesn't exist.


peter-b:
@livecode-vulcan review ok 490c43a

livecode-vulcan:
💙 review by @peter-b ok 490c43a3f3c714d707ec8246ed4e5eacd08eb519

PR #666, opened 2015-11-17
[[ Frame Behavior ]] Multiple issues with the frame behavior

### LINKED ISSUE #581: In narrow project browser, footer tools are missing with no hint
When the project browser is too narrow for all the footer tools to be shown, there is no hint that some tools are missing from the footer.

This screenshot illustrates the issue; there are 5 tools that can't be seen (pushed off the right-hand side) but there's no way to tell!

screen shot 2015-10-23 at 10 18 36

See also: #580

### LINKED ISSUE #593: Tab bars don't cope well with narrow windows
Particularly when in label mode (but it happens in icon mode too), the headers on the IDE palettes can end up with the tabs under the action buttons on the right hand-side.

We could make them wrap to a new line (like Windows Tab Controls do).

We could make little arrows appear left and right to scroll them (I think Safari did this for a while - or Chrome).

We could make a popup appear at the end with any which 'fall' off.

I'm not sure which is the best - there might be other options to.

screen shot 2015-10-23 at 14 02 34

livecodeali:
Closes #593, closes #581

This fixes a number of issues with the frame behavior, and implements some sensible enhancements at the same time:
- Make general generation/layout/resizing more economical
- Add default preference options in the correct place, and for all palettes
- Ensure the palette actions widget always fits and doesn't overlap itself
  (requires https://github.com/livecode/livecode/pull/3289)
- Store and restore some per-palette preferences correctly
- Add PI labels preference to PI stacks
- General cleanup of unused code


peter-b:
@livecode-vulcan review ok 0a51ca1

livecode-vulcan:
💙 review by @peter-b ok 0a51ca1e069b0b97e3f236bca212873044b86013

PR #669, opened 2015-11-18
[[ Bug ]] Icons should be default display style, if no pref set

livecodeali:
sNavStyle is empty if no pref is set, and if the child stack doesn't fix its style. As the message box now doesn't fix its style to allow user pref, it uses the default setting, which was labels rather than icons.


peter-b:
@livecode-vulcan review ok 54b0279

livecode-vulcan:
💙 review by @peter-b ok 54b02795897adbb903ee329c1b63cd4a55a35b19

PR #670, opened 2015-11-18
[[ Tools ]] Make sure tools palette opens at correct size

livecodeali:
The formattedRect of group "contents" seems to be incorrect, so the tools palette opens too tall.

Using the would-be placement of the next tool icon fixes the issue.


peter-b:
@livecode-vulcan review ok 8b9556c

livecode-vulcan:
💙 review by @peter-b ok 8b9556c80d9b314fc8ac5f272fe2aeec31d24fba

PR #681, opened 2015-11-21
[[ Geometry ]] Improve robustness of geometry setting

livecodeali:
In particular, we no longer rely on `this card` and instead derive
the correct card reference from the geometry object.


livecodesebastien:
@livecode-vulcan review ok cdac480

livecode-vulcan:
💙 review by @livecodesebastien ok cdac4806a9cfb2ec656886e2891ef72268231607

PR #683, opened 2015-11-24
[Bug 16464] Prevent script execution abort when resizing stack

livecodepanos:


livecodeali:
Instead of that whole clause:

```
    if there is a stack "message box" then
       ...
    end if
```

you can just call `revSetActiveStacks`.

This will send `ideActiveStacksChanged` to all stacks that have subscribed to this message.

livecodeali:
Hmm, actually that will only work if the active stack has actually changed. I wonder what this was used for...

livecodeali:
Ok, I'm pretty sure that whole clause

```
 if the environment is "development" then
...
end if
```

can be deleted!

Nothing should need to manually update the active stacks anymore, and the previous section is just to make the old property inspector work - the new property inspector detects property changes without having to be manually updated.

livecodeali:
@livecode-vulcan review ok 01de046

livecode-vulcan:
💙 review by @livecodeali ok 01de046ac355f1c455c4740f2df0a41962f68e36

PR #691, opened 2015-11-25
[[ Bug 16483 ]] Fix graphic effect and gradient popup

livecodeali:
These had been broken by the multiple object conflicted value support
which requires values of an editor to be an array keyed by the long id
of the target object.

This is fixed by just adding a 'popup' pseudo object to the values as
set by the editors.

There was also an issue with a lack of lock messages causing the creation of a button to lead to `ideSelectedObjectChanged`, with fairly disastrous consequences for the property inspector.


peter-b:
@livecode-vulcan review ok fe88455

livecode-vulcan:
💙 review by @peter-b ok fe88455d6c979f444bd9afde739199f31609406e

PR #708, opened 2015-12-03
[Bug 16122] Make sure the default backcolor of a fld is white

livecodepanos:


runrevmark:
I'm not sure this is correct. Themed field background colors are fixed - typically white. The problem I think is to do with some of the theme work that has been integrated. If you check pre 8 then fields which use the system theme always have a white background regardless of property settings (at least on Mac).

livecodepanos:
Closing this PR since the bug is probably in the engine code

PR #710, opened 2015-12-06
Support for test stack reuse

montegoulding:
At the moment the extension builder creates a new stack each time the
user tests a widget etc which can be annoying if you want to script some
handlers or just add a few buttons to set properties. This commit adds
support for a stack named test.livecode to be placed in the extension
directory which if found will be loaded instead of creating a new stack.


livecodeali:
@montegoulding Thanks for this, it is a good idea. Could you add a release note in the notes directory please?

livecodeali:
@livecode-vulcan review ok 6f4d23d

livecode-vulcan:
💙 review by @livecodeali ok 6f4d23dd46411e5b4ee4ac8b05778788feec1f48

PR #723, opened 2015-12-08
[[ Dictionary ]] Allow setting of currently displayed API entry from IDE

livecodeali:
The dictionary html has been reworked so that display of an entry is always triggered by the IDE


peter-b:
@livecodeali What happens on platforms where the browser widget isn't available (i.e. the dictionary stack doesn't work)?

livecodeali:
Good point. What a pain.

peter-b:
@livecodeali I'm worried that "Show API" might not be a very beginner-friendly menu entry.  What about "Show Documentation" instead?

peter-b:
@livecode-vulcan review ok 01d70ce

livecode-vulcan:
💙 review by @peter-b ok 01d70ce7247a095d96acdd759ab5707460a45ad5

PR #724, opened 2015-12-11
[Bug 16585] Make sure we remove the stack from memory

livecodepanos:


livecodeali:
Hi Panos, I think the issue here is that tMainstack was not stack reference but just a short name, so it needs to be 
`__ideRemoveFromMemory the long id of stack tMainstack` 

Using pStackID will only delete the target substack, but the previous behavior of 'close and remove from memory' is for the deletion to happen on the mainstack.

peter-b:
@livecode-vulcan review ok 9a7336a

livecode-vulcan:
💙 review by @peter-b ok 9a7336a299346cf48906f0470df96ca767f110a0

PR #725, opened 2015-12-11
[[ Bug 16582 ]] Extend the ide docs api

livecodeali:


peter-b:
@livecodeali Are these block comments supposed to be lcdoc formatted? If so, should they use `/** */`?

Otherwise,
@livecode-vulcan review ok 57d72c6

livecode-vulcan:
💙 review by @peter-b ok 57d72c629a0eef26c6739665d7d9c36cb08b327a

livecodeali:
Heh, good point :-) I guess I'll fix them once they are actually being parsed.

PR #728, opened 2015-12-15
[[ Guides ]] Reorganise and improve guides

livecodeali:


livecodeali:
These are probably the biggest two 'TODO's:
http://quality.livecode.com/show_bug.cgi?id=16610
http://quality.livecode.com/show_bug.cgi?id=16609

PR #734, opened 2015-12-17
[[ Project Browser ]] Allow use of the PB without affecting current selection or focus

### LINKED ISSUE #693: Clicking on cards in the Project Browser fails to navigate to them
LiveCode 8.0.0-dp-10 on Windows 10
- Create a new stack
- Create several cards
- Open the Project Browser
- Click on the cards

Expected behaviour: you can navigate between the cards

Actual behaviour: the current card doesn't change

### LINKED ISSUE #694: Problems copying cards in Project Browser
LiveCode 8.0.0-dp-10 on Windows 10.
- Create two stacks "StackA" and "StackB"
- Make "StackB" a substack of "StackA"
- Add some cards ("Card1", "Card2", "Card3" to "StackB")
- Select "Card1" in the Project Browser
- Right-click on it in the Project Browser and "Copy"
- Select "StackA" in the Project Browser and "Paste Objects"

"Card1" is successfully copied into "StackA"
- Select "Card2" in the Project Browser
- Right-click on it in the Project Browser and "Copy"
- Select "StackA" in the Project Browser and "Paste Objects"

"Card1" gets successfully copied into "StackA" **again**.

The Project Browser shows "StackA" as containing two cards with the same name _and_ ID.  It doesn't appear to be an updating bug because if you close and reopen the Project Browser then the problem persists.

### LINKED ISSUE #714: Project Browser expanding a group opens the stack with group selected
I'm not sure if this is a bug or by design. It makes it impossible to use the project browser to navigate to a grouped control and use the contextual menu on it without opening the stack. I found this in selectObjects:

```
## If a card or stack is selected we don't want to go to the card
## The user should be able to highlight or expand a card or stack without going to it
if word 1 of tObjectList is "stack" or word 1 of tObjectList is "card" then
    exit selectObjects
end if
```

I think the same is true of groups but I also think there is something odd to the design if some objects behave differently to others. Wouldn't it be better to only open the object if the user double-clicks or add a go button and to do the same thing for every object type? Perhaps on selection in the PB select the object if it is on a currently open card but otherwise require some other action to open the card.

livecodeali:
The PB now only navigates/selects when the item is double-clicked.
- Closes #714
- CLoses #693 (navigates successfully on double click)
- Closes #694

### DIFF
--- a/Toolset/palettes/project browser/behaviors/revideprojectbrowsercontainerrowbehavior.livecodescript
+++ b/Toolset/palettes/project browser/behaviors/revideprojectbrowsercontainerrowbehavior.livecodescript
@@ -217,17 +217,13 @@ on mouseUp
          edit the script of the cBehaviorLongID of me
          break
       default
-         pass mouseUp
-   end switch
+         pass mouseUp      
+   end switch   
+   --unlock messages
 end mouseUp
 
 on mouseDoubleUp 
-   if the short name of the target is "name" then
-      set the textColor of field "name" of me to revIDEColor("text_1")
-      CreateFieldEditorForField the long id of field "name" of me, "name"
-   else
-      goToObject the cObjectLongID of me
-   end if
+   goToObject the cObjectLongID of me
 end mouseDoubleUp
 
 on CloseFieldEditor pFieldEditor, pRow, pKey, pClosingTriggeredBy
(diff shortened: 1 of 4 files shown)


peter-b:
@livecode-vulcan review ok f472221

livecode-vulcan:
💙 review by @peter-b ok f472221b8192843fb9f12b572ceb50eceda61ef7

PR #740, opened 2015-12-20
now using LC8 setup filename

mwieder:
Since the linux setup filename has changed in LC8, a couple of scripts that have the name hard-coded need to be changed to match it. If we have to hard-code it (I'd still love to have global constants some day) I'd rather make this a script constant and use that, but I'd still have to do that in each of the two scripts, so I didn't bother.

### DIFF
--- a/Toolset/home.livecodescript
+++ b/Toolset/home.livecodescript
@@ -267,9 +267,12 @@ on revInternal__openStack
                put quote & tCommand & quote && "install -update" into tCommand
                break
             case "Linux"
-               --put item 1 to -3 of the filename me into $PATH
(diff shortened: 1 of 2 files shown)


livecodeali:
Thanks Mark. Yes, it's slightly awkward too because the installer obviously runs without the IDE, so we will always need to change this in at least two locations when it changes. Hopefully it won't change very much anymore though!
@livecode-vulcan review ok 0126260

livecode-vulcan:
💙 review by @livecodeali ok 01262604f070cb630f8bee67fb84758fc6e83f74

PR #741, opened 2015-12-21
Unsubscribe revtools

mwieder:
Revtools.livecodescript never unsubscribes from four messages when it closes, and this causes problems for replacement plugins. This patch symmetrically unsubscribes from the messages that are subscribed on openStack. This also has the effect of fixing http://quality.livecode.com/show_bug.cgi?id=15087


livecodeali:
@mwieder There appear to be some additional dictionary files in the pull request.

mwieder:
I have no idea where those files came from. Is that a result of running the engine/IDE after building it? At any rate, I dumped them and pushed again.

livecodeali:
They're dictionary generated files but they should be `.gitignore`d

mwieder:
Interesting. I do see them in my .gitignore file. No idea why they weren't being ignored.

livecodeali:
It looks like this commit https://github.com/mwieder/livecode-ide/commit/7152a2f0c597841d93c71ff2eb29b8805aadd9e1 is in your develop branch. You should perhaps hard reset and force push it to get rid of it.

I'll close this one as it will be superceded by https://github.com/livecode/livecode-ide/pull/742

mwieder:
I don't know how that commit got in there... possibly trying to avoid build problems here, but it still seems like the .gitignore file should have taken care of it. At any rate, I did a hard reset and pushed, so I think 742 should be good now.

PR #742, opened 2015-12-21
Unsubscribe on closestack

mwieder:
Following on the heels of fixing the subscribe/unsubscribe mechanism for the tools palette, here are similar fixes for three other stacks.


livecodeali:
@mwieder Similarly, some extra files here.

livecodeali:
@mwieder Actually tools, message box, extension builder, extension manager, inspector and dictionary all use the palette behavior, so the unsubscribe all there fixes this for all of those. So if you could remove those extra files and the change to the message box behavior, then this will cover all those situations and supercede https://github.com/livecode/livecode-ide/pull/741.

mwieder:
Cool. Inheritance is wonderful. Changes pushed.

mwieder:
I did a hard reset and pushed, so I think 742 should be good now.

livecodeali:
Thanks Mark!
@livecode-vulcan review ok 771f66a

livecode-vulcan:
💙 review by @livecodeali ok 771f66a6e31c1a42ea88ac369e08c56b5f487f2a

PR #744, opened 2015-12-29
[[ Bug 16440 ]] Unsubscribe from messages when stacks close

livecodeali:


peter-b:
I think these changes relate to [bug 16640](http://quality.livecode.com/show_bug.cgi?id=16640) rather than bug 16440!

mwieder:
Heh.

PR #745, opened 2015-12-30
[[ Cleanup ]] Remove unwanted custom properties from script editor

livecodeali:
Traversed the objects in the script editor stack setting their cRevGeneral["script"] and cRevGeneral["bookmarks"] to empty, as these sometimes contained needless amounts of data (for example, in one case, the entire script of revIDELibrary).

Drops the file size down from 3.2MB to 325KB


peter-b:
I'm not sure how to actually review these changes!  However, they make perfect sense and slim down the IDE a lot, so I think we should go ahead and merge them.

@livecode-vulcan review ok 70effe1

livecode-vulcan:
💙 review by @peter-b ok 70effe1353af6cb4590536b4aacd17f83c7c8243

PR #759, opened 2016-01-07
[[ Dictionary ]] Ensure all dictionary data modifications occur outwith the application bundle

livecodeali:
Previously the IDE would try to build dictionary data in /Documentation/html_viewer/resources/data, which is a problem for the installed application on Windows and Linux (as the location is not necessarily writable) and means that the Mac IDE can't be signed.

This pull request changes this so that all the data is built and modified in an appropriate 'Application Support' folder, or equivalent.


peter-b:
@livecode-vulcan review ok fce6f83

livecode-vulcan:
💙 review by @peter-b ok fce6f834f38973f1a7ae40f784ad528fd055d66e

PR #760, opened 2016-01-07
Don't set an explicit font on the home stack on OSX

livecodefraser:
The theming code knows how to handle it and setting it on the home stack just causes unnecessary differences between IDE behaviour and standalone behaviour.

Needs to be merged after https://github.com/livecode/livecode/pull/3438


livecodeali:
@livecode-vulcan review ok ff6018d

livecode-vulcan:
💙 review by @livecodeali ok ff6018d9768dd6ad2a39688bbba57354de6f073f

PR #768, opened 2016-01-11
[Bug 15970] Correct some uses of revised "save" command

peter-b:
- Change some other places where the `save` command is used in the IDE
  to use either `with format _` or `with newest format` as appropriate
- Don't pass an empty to save as a stack file format
- Use `revInternal__SavePreferences` instead of saving the
  "revpreferences" stack directly


livecodeali:
Looks good :-)
@livecode-vulcan review ok de6a3a8

livecode-vulcan:
💙 review by @livecodeali ok de6a3a8ff23c41502b61f31cdc9fd04c7bc703b8

PR #772, opened 2016-01-12
[[ PI ]] Fix some things that will be noticeable when native theming is improved

livecodeali:
Namely:
- The field in the string editor should set its effective rect rather than its rect
- The string and boolean editors shouln't have explicit font sizes set


peter-b:
@livecode-vulcan review ok 68f67ad

livecode-vulcan:
💙 review by @peter-b ok 68f67adef7f6008ed18cb6c94a9603fd512c55e2

PR #777, opened 2016-01-12
[Bug 16711] Disable "Stacks" page for Android, iOS and Emscripten

peter-b:
None of these platforms work with the "Stacks" page.  Previously, the
stacks page was only disabled if "iOS" was selected as a deployment
platform.


livecodeali:
If this does what it says it does then it's fine :-)
@livecode-vulcan review ok b9b0690

livecode-vulcan:
💙 review by @livecodeali ok b9b06900c9525e509b172aba97a8dbbcf5fc1f08

PR #787, opened 2016-01-14
[[ PI ]] Lock inspector when a control is selected thereon

### LINKED ISSUE #709: Property Inspector: Not all objects in the list provided exist
I tried to inspect the properties of the "Style" control of card 1002 of stack "revPropertyInspector1" using LC8.0-dp10 on 64-bit linux Mint 17 and was greeted with the following message in the message box:

ERROR: Not all objects in the list provided exist
line: 4426
handler: __revIDEPropertiesInfoOfSection
script object: stack "/opt/livecode/livecodecommunity-8.0.0-dp-10.x86_64/Toolset/libraries/revidelibrary.8.livecodescript"

livecodeali:
Closes #709

### DIFF
--- a/Toolset/palettes/inspector/revinspector.livecodescript
+++ b/Toolset/palettes/inspector/revinspector.livecodescript
@@ -300,11 +300,24 @@ end ideNameChanged
 
 # Sent by the IDE when the selection changed
 on ideSelectedObjectChanged
+   local tObjects
+   put revIDESelectedObjects() into tObjects
+   
+   local tStack
+   put revIDEStackOfObject(line 1 of tObjects) into tStack
+   
+   # If the object is on an inspector stack, lock it
+   if the short name of tStack begins with kPropertyInspectorPrefix then
+      if the cLocked of tStack is not true then
+         dispatch "actionSelected" to tStack with "inspectorLock"
+      end if
+   end if
+   
    local tStackName
    repeat for each key tKey in inspectorList()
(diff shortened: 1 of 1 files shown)


peter-b:
@livecodeali It feels like this should be pluggable in some way to facilitate development of third-party inspector-like stacks... However, I can't see any obvious way to implement that (apart from maybe implementing pre/on/post phases for IDE messages?)

@livecode-vulcan review ok 99d9e89

livecode-vulcan:
💙 review by @peter-b ok 99d9e8922a3096892690c7bfe817f6d50a5f5cce

livecodeali:
@peter-b Yes, I've been thinking about this recently as it has arisen a couple of times in various contexts.

There are at least two IDE components (PI & Script Editor) that need to manage multiple instances. I think there's a lot of work to do to make a friendly API for these things. At the moment I think we just have to allow people to replace the manager itself rather than the instances it manages.

It also came up when thinking about the menubar - single-instance palettes should probably have menubar toggles, whereas that's not really the right thing for the PI or script editor, as they should be buttons adding a new instance.

peter-b:
@livecodeali Since you've (clearly) already put some thought into this issue, could you please do a brain dump into an issue (or BZ ticket) and mark it for future-8.1?  I'd rather it didn't get forgotten about...! In the mean time I think it's okay to merge these changes.

livecodeali:
@peter-b Will do :-)

PR #804, opened 2016-01-22
Bugfix binary stack layout

livecodeali:
Almost all of these are just fixing the widths of fields and containing groups.
The script lines labels in the project browser now have themeClass "label" to make them look correct.


livecodeali:
I'm not sure there is much point in reviewing this.

PR #814, opened 2016-01-26
[[ Bug 16751 ]] Ensure enum editor always displays current value

livecodeali:


peter-b:
@livecode-vulcan review ok 6a6e186

Maybe the problem here is that the font property isn't _really_ an enum?  This seems like a workaround for the absence of a "proper" font editor (but that doesn't make it invalid).

livecode-vulcan:
💙 review by @peter-b ok 6a6e186fd8a17749409ccfef01624b437faafcdb

PR #829, opened 2016-01-28
dictionary: Make crossrefs in description blocks look like normal text

peter-b:
Some dictionary entries have a _lot_ of cross references in their
description blocks, which can be extremely visually noisy.  Since all
the cross references in the description are already listed in the
dictionary entry's references section, it's okay to make the links in
the description block look like normal text until mouseover - it
doesn't hinder discoverability but it does make it easier to read.

### DIFF
--- a/Documentation/html_viewer/css/lcdoc.css
+++ b/Documentation/html_viewer/css/lcdoc.css
@@ -124,3 +124,13 @@
     display:block;
   }
 }
+
+.lcdoc_description a.load_entry {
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok e34a9e3

livecode-vulcan:
💙 review by @livecodeali ok e34a9e38b881123d3a45003db648e685aa26c9f4

PR #840, opened 2016-01-29
[[ PI ]] Beef up behavior PI editor

livecodeali:
The behavior editor now has an object selection menu from which a
suitable behavior object can be chosen. In addition to the 'edit script'
button which opens the script editor for the chosen behavior object,
there is now also an "Inspect object" button which navigates to the
behavior object and causes it to be selected, whilst locking the current
inspector.

This PR provides a general function 

```
function revIDEObjectSelectionMenu pTargetObjects, pTypeFilter
```

in the IDE library, which is essentially a refactored form of the inspector's 
'inspect' menu but with an added object type filter. This is used by the 
behavior editor to provide a selector for buttons and stacks.


runrevelanor:
@livecode-vulcan review ok 1ad212e

livecode-vulcan:
💙 review by @runrevelanor ok 1ad212eb46a3147beb856c0f69af7a6100948913

livecodeali:
screen shot 2016-02-17 at 15 16 16

PR #851, opened 2016-02-03
[[ PI ]] Use formattedheight for height of all fields in editors

livecodeali:
Unfortunately this results in the fields being too high. So we
can do various margin fudges and so on, but ideally this is something
we could sort out at the engine level.


livecodeali:
All field defaults + setting height to formattedHeight:
screen shot 2016-02-03 at 10 39 21

peter-b:
That looks to me like the formatted height is being computed using a text baseline that's midway down the field.

BerndN:
@livecodeali 
using the effective textHeight + one third of the effective textheight seems to work for a variety of font sizes and fonts. Tested with one-liner and set the scroll to 100. Then tabbing into the field and starting to type does not shift text.

set the height of field x to trunc ( the effective textHeight of field x +( the effective textHeight of field x / 3.3333) )

is not as tall as using formattedHeight = waste of space
Using formattedHeight of line 1 of field x shifts text down.

runrevmark:
The problem here is that the metrics you get from a field are 'fudged' a fair bit... We should be able to sort this out properly in the engine now, resorting to the old 'broken' metrics if the field's theme property is 'legacy'.

Let us call the actual specified rect of a field the ControlRect.

We then have the VisibleRect which is the ControlRect reduced by borders and scrollbars.

We compute the LayoutWidth as the widget of VisibleRect - leftMargin - rightMargin.

If the field is wrapping, then the text is wrapped to the LayoutWidth.

In terms of the metrics of the lines themselves (i.e. lineHeight / ascent / descent / leading) then we are probably best to copy the semantics from CSS2/3 (https://www.w3.org/TR/CSS2/visudet.html#line-height). The textHeight property would become the 'lineHeight' CSS property.

The DocumentRect is then the rect bounding the (wrapped) text but including the specified margins.

The RectOfLine is the bounding-box of that paragraph, including any spaceAbove/spaceBelow properties.

The RectOfChar is the bounding-box of the portions of the lines the chunk covers - including leading.

This should mean that the height you need to set a field to in order to make sure it doesn't clip the text in any way would be the formattedRect of the field appropriately expanded by the borderWidth and scrollbars.

BerndN:
@runrevmark 
Thanks, that would be most welcome since it takes a lot of "guessing" and workarounds out of this.

livecodeali:
This will be ok when this is merged: https://github.com/livecode/livecode/pull/3589

peter-b:
@livecode-vulcan review ok 47cc569

livecode-vulcan:
💙 review by @peter-b ok 47cc56989f1966d008249e16dceebc7b9bd5cfd1

PR #854, opened 2016-02-03
[[ Widget Default Scripts ]] Add mechanism for creating, packaging and using widget default scripts

livecodeali:
The default script must be in a script-only stack called 

```
<widget_identifier>.__DefaultScript
```

in a file called `defaultscript.livecodescript` in the widget's `support` folder.

The extension builder stack has an additional field for displaying the default script and a button for editing it.
screen shot 2016-02-03 at 15 34 14


Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -998,5 +1006,46 @@ private function __revIDEDeveloperExtensionAddSpecifiedFilesToPackage pSource, p
...
+   put pFolder & "/support/defaultscript.livecodescript" into tDefaultScript
+   
+   local tExpectedName
+   put sExtensionDetailsA[pFolder]["id"] & ".__DefaultScript" into tExpectedName
+   if there is a file tDefaultScript then
+      try
+         # If we have a valid stack, open it and edit the script
+         open stack tDefaultScript
+         put the long id of stack tDefaultScript into tTargetStack
+         revIDEEditScriptOfObject tTargetStack
```
peter-b, line 1032: Is it necessary to add a check that `the name of tTargetStack is tExpectedName` here?

livecodeali, line 1032: Maybe... I wondered if it was overkill at that point. Perhaps the best thing to do would be to warn that the name is not the expected name and give the option to rename automatically?

peter-b, line 1032: Sounds good to me.

peter-b:
I don't think that fix should block this going in.

@livecode-vulcan review ok b602330

livecode-vulcan:
💙 review by @peter-b ok b60233011e7adec516a6f0064799d15ccda58abb

PR #888, opened 2016-02-11
Read only strings should still be able to be selected in the UI

trevordevore:
Currently a read only string (e.g. widget <type>) field in the PI is disabled. This means you can't select the text and copy it. This commit sets the locktext property and the textcolor property rather than setting the enabled property.


livecodeali:
@livecode-vulcan review ok f17097b

livecode-vulcan:
💙 review by @livecodeali ok f17097bd8b22592dc6b4f78279f192f4de170ab4

PR #914, opened 2016-02-17
Ensure cached script editor fields are hidden

### LINKED ISSUE #910: Turning on Show Invisible Objects makes the Script Editor unusable
Turning on Show Invisible Object shows an offset duplicate of the code, this makes it very hard to use.

A "context" button is also shown in the Menu bar.

se_invisibleoff

se_invisibleon

peter-b:
When the engine was set to show invisible objects, the script editor
became unusable because the cached editor fields would be displayed on
top of the "real" editor field.

This patch ensures that cached fields are sent to the bottom layer and
uncached fields are brought to the front.

Closes #910.


livecodeali:
@livecode-vulcan review ok 9577b5f

livecode-vulcan:
💙 review by @livecodeali ok 9577b5f642ff57d7137b645c55dbc2a6fc5e9bb3

PR #937, opened 2016-02-24
[[ PI ]] Show object select menu on mousedown in behavior editor

### LINKED ISSUE #916: PI: Behavior selector should is a bit odd
The behavior selector should popup on mouseDown rather than mouseUp. I'd also be keen to be able to drag a line out to the button, stack or project browser similar to Interface Builder but I'm guessing that would be rather interesting to implement in a multi-window environment. It's worth noting that the reason I thought of the latter is because of the mouseDown highlight and no popup made me thing it was how it worked and it wasn't until I finally mouseUp'ed on on the widget that I realised there's a popup.

livecodeali:
Closes #916

### DIFF
--- a/Toolset/palettes/inspector/editors/com.livecode.pi.script.behavior.livecodescript
+++ b/Toolset/palettes/inspector/editors/com.livecode.pi.script.behavior.livecodescript
@@ -85,16 +85,22 @@ on editorEditScript
    end try
 end editorEditScript
 
-on mouseUp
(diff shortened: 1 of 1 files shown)


peter-b:
@livecode-vulcan review ok 27f591e

livecode-vulcan:
💙 review by @peter-b ok 27f591e02a80f35ec2d3745dbff24a511d7600fb

PR #940, opened 2016-02-24
[Bug 16992] Make sure handling "saveStackRequest" cancels IDE save

peter-b:
According to the dictionary, handling the `saveStackRequest` message
(and failing to pass it) should cancel the `save` command.

The IDE calls the `save` command with messages locked, which prevents
the current stack from receiving the normal `saveStackRequest`
message.  The IDE therefore synthesizes one.

Previously, the IDE ignored whether or not the `saveStackRequest`
message was handled, and continued with the save operation regardless.
This prevented some useful applications of of intercepting the `save`
command (e.g. forcing the on-disk format to a specific stack format
version).  This patch ensures that the IDE save operation is cancelled
if the `saveStackRequest` message is blocked.

### DIFF
--- a/Toolset/libraries/revfrontscriptlibrary.livecodescript
+++ b/Toolset/libraries/revfrontscriptlibrary.livecodescript
@@ -277,23 +277,6 @@ on preOpenStack
    if not gREVSuppressMessages or (gREVSuppressMessages and revOKTarget()) then pass preOpenStack
 end preOpenStack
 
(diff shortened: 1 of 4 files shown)


livecodeali:
Looks good to me.
@livecode-vulcan review ok 902432c

livecode-vulcan:
💙 review by @livecodeali ok 902432c3e2d6f07166b151289b0e9b7f272d64af

PR #946, opened 2016-02-26
[[ Flip/Rotate ]] Fix flipping and rotating from menubar

livecodeali:
The rotate command is especially tricky as the form `rotate <angle>`
to rotate the selected object means that there is a parsing ambiguity
if we allow arbitrary image chunk expressions for `rotate <image> <angle>`.
So we have to force the command to include the object type `image` explicitly.

The flip command was fixed in 6.7.8 to accept any image chunk expression, as
it doesn't suffer the same ambiguity as rotate - which axis to flip on is
either "horizontal" or "vertical" - it does not evaluate an expression.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -8039,10 +8039,13 @@ on revIDEFlipImages pImages, pOrientation
    
    lock screen
    repeat for each line tImage in pImages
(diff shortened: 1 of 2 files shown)


peter-b:
@livecode-vulcan review ok 9c7af18

livecode-vulcan:
💙 review by @peter-b ok 9c7af18a8ea75aa653c5290255fbe8c1693b8a69

PR #965, opened 2016-03-04
[[ PI ]] Fix several issues with the gradient editor

livecodeali:
- Don't pop up editor when right-clicking
- Don't pop up color dialog when no stop is selected
- Use current stop color as preselect when popping up color dialog

This PR has slightly less in it than it looks, because some code that used
to be in a group script is now in the script-only behavior


peter-b:
@livecode-vulcan review ok 899b7b2

peter-b:
@livecodeali I think this might need a release note.

peter-b:
@livecodeali Ah -- I see that we still can't handle GitHub bug numbers in V1-format release notes. Oh well!

@livecode-vulcan review ok 899b7b2

livecode-vulcan:
💙 review by @peter-b ok 899b7b2f992710ba28716c8b9395b00de9c0ccc1

PR #993, opened 2016-03-14
Allow keyboard navigation of answer dialogs on Linux again

livecodefraser:
When the answer dialog layout code for Linux was changed to right-align the answer dialog buttons, as on OSX, it also meant that the code path disabling keyboard navigation was also run on Linux. This has been split out from the layout code to restore navigation on non-OSX platforms.


livecodeali:
Can you post a gist of the diff please?

livecodefraser:
Card script of the only card in revanswerdialog:

```
--- before.txt  2016-03-29 12:07:14.371172035 +0100
+++ after.txt   2016-03-29 11:47:42.014788855 +0100
@@ -290,13 +290,20 @@
             subtract the width of button i + 12 from tStartEdge
          end if
          set the bottom of button i to the height of this stack - 12
-         set the traversalOn of button i to true
       end repeat
    end if
    if the platform is "MacOS" then set the top of this stack to round(item 4 of the screenRect/5)
    set the default of button 3 to true
    #debugLog "DONE"

+   -- Provide keyboard navigation on Linux and Windows but not OSX
+   if the platform is not "MacOS" then
+      -- Make sure that we don't make the icon buttons navigable
+      repeat with i = (tNoOfButtons + 2) down to 3
+         set the traversalOn of button i to true
+      end repeat
+   end if
+   
    --get the rect of field 1 of me
    --put the width of this stack - 12 into item 3 of it
    --set the rect of field 1 of me to it
```

livecodeali:
@livecode-vulcan review ok 1b3349a

livecode-vulcan:
💙 review by @livecodeali ok 1b3349a4fcc59d0acb2612fd02e350e5c27619d4

Issue #992, opened 2016-03-14
Background colour for pending messages, front and back scripts and stack in use


runrevelanor:
I think it might look better if the background colour for these panes was white rather than the background colour for the stack. When there is nothing in the list it looks like it hasn't displayed properly, rather than it being an empty list, to me at least.

Issue #948, opened 2016-02-29
Message Box: can't copy from multiline results


runrevelanor:
You can't select from the results field on the multiline pane. Clicking and dragging anywhere in the results field activates the top/bottom section adjuster.

PR #1010, opened 2016-03-15
Improve the multi-line mode of the Message Box

runrevelanor:
- Only allow the separator to resize the 2 panels
- Allow text to be selected and copied from the results field

Change the background color for the pending messages, front and back scripts
and stacks in use to white (issue #992)


runrevelanor:
Closes: #948 
Closes: #992

livecodeali:
Can you add a `constant kSeparatorSize = 10` or something like that, and replace the 5s with `kSeparatorSize / 2` - that way we can change it easily in one place if it needs changing, and the 5s are less mysterious.

livecodeali:
@livecode-vulcan review ok b8396c7

livecode-vulcan:
💙 review by @livecodeali ok b8396c7b99e00a81396838a608bfe8d680066b9f

PR #1033, opened 2016-03-18
[Bug 17158] More uniform implementation of revidemessagehandlerlibrary

peter-b:
Ensure that all handlers in the IDE message handler library are
implemented in the same form:

```
 if it's a user stack then
    send an IDE message
 end if
 pass message
```

This is intended to make it harder to miss out a `pass` when making
changes to the library in the future.


livecodeali:
@livecode-vulcan review ok c07eaa3

livecode-vulcan:
💙 review by @livecodeali ok c07eaa3f98552639e76703f2061011eeabcfbf94

Issue #1026, opened 2016-03-17
Default repeatCount of imported gif


runrevneil:
The default repeatCount for an imported gif is set to "0" in LC8. In previous versions of LC this is initially set to "-1"

runrevelanor:
In 8 images are imported by creating an empty image and then setting the text of it, this means that an image gets the default value for its repeatCount property, however this should be -1 of animated gifs and 0 for other images.

In 7 and earlier images are imported using the import paint from file command, which seems to automatically set the repeatCount.

I haven't updated revIDEImportControl as I am not sure what is more correct.

livecodeali:
We should probably change it back to using the `import paint from file` command - I'm not sure why it was changed in the first place.

runrevelanor:
Will do.

PR #1034, opened 2016-03-18
Import images directly, automatically setting the repeatCount

runrevelanor:
When imorting images using the File menu images will be imported
using the import paint from file command. This automatically sets
the repeatCount property of the new image.

Closes: #1026

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -5879,13 +5879,13 @@ on revIDEImportControl pType, pFileName, pReferenced
    set the defaultStack to tTargetStack
    switch pType
       case "image"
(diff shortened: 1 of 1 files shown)


peter-b:
@runrevelanor In order to avoid this problem recurring in the future, it might be a good idea to add some comments to the code that explain that that's the reason `import paint` is used here.  What do you think?

runrevelanor:
I can do, it is how it was always done before 8 so is not really a change. If this seems like a good place to add a comment I'll add to the revIDELibrary handler that imports controls.

peter-b:
Sounds good to me.

peter-b:
@livecode-vulcan review ok 001dbae

livecode-vulcan:
💙 review by @peter-b ok 001dbae900e75a9228b097db2e2ab2b102cd47eb

PR #1042, opened 2016-03-21
Correctly display password protected stacks in the Message Box

runrevelanor:
Display "pwd" instead of the number of script lines for password
protected stacks used as front scripts, back scripts or library
stacks when displayed in lists in the Message Box.


peter-b:
@runrevelanor There are several places in the IDE that need to get a suitable count of script lines to place in a button -- at least the message box and the project browser.

Does it make sense to take this opportunity to add a suitable helper to revidelibrary?

peter-b:
@livecode-vulcan review ok a13a0cb

livecode-vulcan:
💙 review by @peter-b ok a13a0cb7f98617c4ebded847913da3f1b8014275

PR #1050, opened 2016-03-23
[[ Geometry Manager ]] Ensure default stack is correct when setting geometry

livecodeali:
The object reference in the geometry array is the `name` of the referenced object
rather than the long id, so the defaultStack must be set the the object's stack.


peter-b:
@livecode-vulcan review ok d4aea7b

livecode-vulcan:
💙 review by @peter-b ok d4aea7b92f58cc71443a4a16b3e2d80c2f855e8b

PR #1059, opened 2016-03-24
[[ Dictionary ]] Order filter options alphabetically

### LINKED ISSUE #901: Dictionary: Order tags alphabetically

livecodeali:
Closes #901

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -165,16 +165,9 @@
 		}
 	}
 	
-	function sortedKeys(array)
+	function sortedKeys(obj)
 	{
-    	var keys = [];
-
-		for(var key in array)
-		{
-			keys.push(key);
-		}
-
-		return keys.sort();
+		return Object.keys(obj).sort();
 	}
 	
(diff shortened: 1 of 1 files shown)


Documentation/html_viewer/js/dictionary_functions.js
```diff
@@ -165,6 +165,18 @@
...
+	{
+    	var keys = [];
+
+		for(var key in array)
+		{
+			keys.push(key);
+		}
+
+		return keys.sort();
+	}
```
peter-b, line 171: Arrays in Javascript are like LCB's lists -- it might be clearer if you rename the argument to `sortedKeys` from `array` to `obj`.

Did you try writing this as something like the following?

``` javascript
function sortedKeys(obj)
{
    return obj.keys().sort();
}
```

livecodeali, line 171: That would be much better. Didn't know there was a keys function!

peter-b:
@livecode-vulcan review ok ecbd493

livecode-vulcan:
💙 review by @peter-b ok ecbd49334cdd860f01f67959c45d8ec5c14836f4

PR #1076, opened 2016-03-30
Set the ID of any images created on IDE stacks to a reserved ID

### LINKED ISSUE #860: The Icon Picker doesn't show imported images.
If you import images for use as button icons and they use the icon picker to choose them the images are not shown in the picker.

2016-02-04 15_31_51-program manager

runrevelanor:
The Menubar, Tools Palette and revIDELibrary all use images, the IDs
of some of these images were conflicting with image IDs on user stacks
resulting in the icon chooser showing IDE images rather than user
images. Setting the ID of images on IDE stacks to within the
reserved range when they are created should prevent this conflict.

Closes #860

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -2438,6 +2438,26 @@ function revIDEIconID pIcon
    return the ID of image pIcon of card 1 of stack "reviconsnew"
 end revIDEIconID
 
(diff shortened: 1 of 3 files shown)


livecodeali:
@livecode-vulcan review ok b47398d

livecode-vulcan:
💙 review by @livecodeali ok b47398d391fc51344db921e9b2b03b0687b2c3b0

PR #1081, opened 2016-03-30
Set the theme of the Preference GUI stack to "legacy" on Linux

runrevelanor:
Text in the Preferences GUI is truncated on Linux due to theming
changes and variable native fonts and text sizes. Setting the theme
to "legacy" restores the fixed-font-and-size settings the stack was
originally designed for


livecodeali:
@livecode-vulcan review ok 79dd2fa

livecode-vulcan:
💙 review by @livecodeali ok 79dd2fa1e26a439f08807f1bc90b759e35c1f172

PR #1083, opened 2016-03-30
Update the Project Browser when a control is pasted into a group

runrevelanor:
The Project Browser receives the ideNewControl message when a control
is pasted into a group but the object ID passed with the message
does not include the fact that the new control is grouped. By using
the long ID of the passed object ID the Project Browser correctly updates
with the new, grouped control.


livecodeali:
@livecode-vulcan review ok 6574372

livecode-vulcan:
💙 review by @livecodeali ok 65743727972aebd102d3d6dc00a461a94ad893f5

PR #1084, opened 2016-03-31
[[ PI ]] Use there is not a to determine existence of pi objects

livecodeali:
Exists can take an arbitrary chunk expression, so exists (line x
of tVar) will attempt to evaluate tVar as a container (i.e. field).

Use `there is a` to force evaluation of a line of the variable as
an arbitrary object.


peter-b:
@livecode-vulcan review ok a763c2f

livecode-vulcan:
💙 review by @peter-b ok a763c2f697bec00e2eff59ec89c71f49ffe78ac4

PR #1085, opened 2016-03-31
[[ PI ]] Don't group scrollbar editors

livecodeali:
Just to clarify, the last `false` used to be in the Group column. I've added a tab to move it to the Default column.


peter-b:
@livecode-vulcan review ok bc06db2

@livecodeali I'm concerned that the tab-delimited format for the PI editor definitions are causing unnecessary whitespace-related breakage that's actually really difficult to check for in GitHub diff views.  To what extent is it viable to switch to some more easily-verified format after 8.0 release?

livecode-vulcan:
💙 review by @peter-b ok bc06db24f85b98f8c6404566e82d5a2b0b681566

PR #1086, opened 2016-03-31
[[ Bug 17279 ]] Relativize appropriate stackFile paths

runrevmark:
This patch adds a custom setprop handler for the stackFiles property
of stacks.

The custom setprop handler will relativize any file paths in the new
value relative to the target object in the case where the referenced
file is in the same folder, or in a folder beneath the folder containing
the target object.


livecodeali:
@livecode-vulcan review ok 91adb00

livecode-vulcan:
💙 review by @livecodeali ok 91adb0052586533732db449edf983607019aef42

PR #1091, opened 2016-04-01
Fix Message Box errors when setting background layer

runrevelanor:
In some cases the layer of the "background" group is set when
the stack is not open, this causes an error and is resolved by
attempting to set the layer within a try.

There was also an error the first time the Front Scripts tab was
opened, due to referring to a control in the footer than did not exist,
moving the reference to openCard fixed the error.


livecodeali:
@livecode-vulcan review ok 597eb59

livecode-vulcan:
💙 review by @livecodeali ok 597eb59caeae794a8de692ecd02c1fdccd01e928

PR #1094, opened 2016-04-01
Make sure the OSX and Windows standalone settings are preserved

livecodepanos:
Bug 16689 - Windows icon is lost when switching tabs:
When you are on Windows card of Standalone Settings and you click the "Choose..." button to select an icon, the `propertyChanged` message is sent to the group "Windows". This message is not any more handled in the "Windows" group, since it has been renamed to `WINpropertyChanged`. This patch makes sure that the correct message is sent. Similarly for OS X card.

Bug 17120 - Stack opened with LC 8 loses LC7 standalone settings
Stacks created with LC7 have an empty value for the custom property `cSettingTag` of button MacOSX x86_64 (since 64-bit builds for OSX are not supported in LC7). If a stack created with LC 7 is opened in LC 8, an error was caused when trying to `set the hilite of btn MacOSX x86_64` to the `cSettingTag of button MacOSX x86_64` (i.e. to empty). This patch makes sure that the `hilite` property is set to `false` in that case


livecodepanos:
Binary diff:

https://gist.github.com/livecodepanos/6efe33a184410e3fd5d4f42ff030bdb0

livecodeali:
@livecode-vulcan review ok 8c33adf

livecode-vulcan:
💙 review by @livecodeali ok 8c33adfcafa0655f26e7c559d05ff912bd3a0b74

PR #1110, opened 2016-04-06
[[ PI ]] Fix various issues with PI appearance 

livecodeali:
- Don't expand numeric text fields to fill the width of the PI
- Adjust some min widths
- Left-align option controls
- Put multi-object align tab at the end
- Ensure frame is always correct size
- Remove unused metal and liveResizing properties


livecodeali:
@livecode-vulcan review by @peter-b ok f3a7667

PR #1113, opened 2016-04-07
[Bug 17301] Make sure grouping controls respects the layer order

livecodepanos:
The `selectedObject` property is an ordered list of selected objects - ordered by the order in which they were selected.

The IDE was building a group command containing all the objects in the selected object list, which caused the objects to be grouped _in that order_. In this patch, the IDE uses the simple `group` command (i.e. with no objectList as a parameter) which groups the selection in layer order.


peter-b:
@livecode-vulcan review ok 6c16efc

livecode-vulcan:
💙 review by @peter-b ok 6c16efc924f79eeee97991289b7ed7eeb2fae6b8

PR #1120, opened 2016-04-08
Check for the existence of an object before adding it to the PB

runrevelanor:
In some cases the Project Browser was receiving ideNewControl messages
for objects that had already been deleted, causing a "no such object"
error. The Project Browser now checks the existence of an object
before adding it to the display.

Additionally these IDE errors were being displayed in the Message Box.
IDE errors are now only shown in the Message Box if gRevDevelopment is
true.


livecodeali:
@livecode-vulcan review ok ff0a59d

livecode-vulcan:
💙 review by @livecodeali ok ff0a59d6aafb95e5ed82215f6774e0558b9caeef

PR #1121, opened 2016-04-08
[[ Bug ]] Set default properties of objects correctly

livecodeali:
- Use setters from prop definition files when setting object default props
- Prevent table props from clobbering other props
- Fix error in tests due to using table library


peter-b:
@livecode-vulcan review ok f325984

livecode-vulcan:
💙 review by @peter-b ok f325984b5ded2dea4da76366be88290c24973f42

PR #1124, opened 2016-04-08
Display the Front Scripts in the Message Box correctly on first open

livecodeali:
An error was being thrown when setting the palette footer items when
the Front Scripts card as first opened, this prevented the list
being drawn.


livecodeali:
Just realised this should have gone into develop-8.0.

PR #1128, opened 2016-04-11
Prevent the IDE handling newBackground and deleteBackground messages

### LINKED ISSUE #1127: Grouping controls does not correctly update the Project Browser

runrevelanor:
When a group is created newGroup and newBackground messages are sent,
the newBackground messgage is for compatibility with HyperCard. Both
messages are mapped to an ideNewControl message which was causing the
Project Browser to believe 2 new groups had been deleted and update
accordingly. Removing the newBackground message from the Message
Handler Library ensures only the newGroup message is handler.

The deleteBackground message has also been removed for the same reason.

Closes #1127

### DIFF
--- a/Toolset/libraries/revidemessagehandlerlibrary.livecodescript
+++ b/Toolset/libraries/revidemessagehandlerlibrary.livecodescript
@@ -149,24 +149,6 @@ on newWidget
    pass newWidget
 end newWidget
 
-on newBackground
-   local tTarget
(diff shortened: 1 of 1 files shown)


livecodeali:
@livecode-vulcan review ok 9ddc5f2

livecode-vulcan:
💙 review by @livecodeali ok 9ddc5f226f849b8e02c88720117eb177dbadf457

PR #1134, opened 2016-04-14
Make sure fake fonts do not appear as disabled in the font list

livecodepanos:
Added to the script of group "Script Editor" of card "Script Editor" of stack "revPreferencesGUI", in command "update", after line `sort lines of tFontText`:

```
/* Fake fonts begin with "(", but the menu code interprets a "(" at 
the beginning of the item to mean the item is disabled. 
So, to show a "(" instead of disabling, we need to double it */

repeat with tLine = 1 to the number of lines of tFontText
   if line tLine of tFontText begins with "(" then
      put "(" before line tLine of tFontText
   end if
end repeat
```


peter-b:
@livecode-vulcan review ok a878525

livecode-vulcan:
💙 review by @peter-b ok a8785252e0b3cdee3aa1ec87cd957fa92d2d4e42

livecodepanos:
This needs a bit more work, since on Linux the font list is displayed as expected _without_ this patch. We probably have to check if the platform is MacOS first.

Moreover, this problem appears in the menuBar as well (Text -> Font) but in this case the list has disabled items on _both_ OSX and Linux. Marked as WIP for now.

livecodepanos:
Ok, to sum up:

This problem appears in two places:
1. in revPreferences
2. in revMenubar

(1) in revPreferences it affects only OS X
(2) in revMenubar it affects all platforms

After offline discussion with the dev team, it was decided that:
1. Issue (1) is an engine bug and should be dealt with separately
2. Issue (2) can be fixed by adding an extra "(" in the beginning of the items beginning with "("

So I am closing this PR and I will submit a new one for the menubar issue

PR #1135, opened 2016-04-14
[Bug 17404] UI fonts do not appear correctly in the menubar font list

livecodepanos:
Fake fonts begin with "(", but the menu code interprets a "(" at the beginning of the item to mean the item is disabled. So, to show a "(" instead of disabling, we need to double it


livecodefraser:
@livecode-vulcan review okay 21b78ce

livecode-vulcan:
💙 review by @livecodefraser ok 21b78ce5324f235b6a2c1e642fec9916c5f47410

PR #1137, opened 2016-04-15
Show the correct state of the Select Grouped icon in the Menubar

runrevelanor:
On opening the Menubar sets the state for the Select Grouped icon
based on the value of the selectGroupedControls property, Previously
the icon was set based on not the selectGroupedControls, causing the
icon to appear in the wrong state.

The default icon for Select Grouped is off and the highlighted icon is
on, the images for these states were inverted, the image files have
been renamed so the correct icon is used for each state.


peter-b:
@runrevelanor Can you add a release note please?

peter-b:
@livecode-vulcan review ok dde7ea4

livecode-vulcan:
💙 review by @peter-b ok dde7ea4243d5112ac0e050f0bc21fac4712b4294

PR #1148, opened 2016-04-25
[[ Bug 17503 ]] Ensure custom control project browser long IDs are valid

livecodeali:


peter-b:
@livecodeali I'm feeling a bit stupid because I can't figure out how the change fixes the bug.  As far as I can tell the "before" and "after" code should do exactly the same thing! `tControlArray["long id"]` isn't accessed in the code between the old location it's set and the new location it's set, and there are no ways for control flow to execute the function (except via an error being thrown).  I'm clearly missing something here; could you explain please? 😄

livecodeali:
@peter-b There's a line in the middle there that does

```
put "Custom control" into tControlArray["type"]
```

so this ensures the long id is generated before the type is changed.

peter-b:
@livecodeali Thank you!

@livecode-vulcan review ok a4d5f23

livecode-vulcan:
💙 review by @peter-b ok a4d5f23279d0fb96ee6dda539af52568498bef34

PR #1156, opened 2016-04-29
[Bug 17354] Make sure msg box always remembers previous location

livecodepanos:
When the msg box was opened/closed, the `moveStack` handler of stack `revMessageBoxBehavior` should be called to set the position of the stack as a custom property. Instead, the `moveStack` handler of stack `revIdeMessageHandlerLibrary` was called, which was changing the `top` of the msg box under the condition:

```
if pTop < the bottom of stack tMenuBar and the right of stack revTargetStack() < the right of stack tMenuBar + 5 then
   set the top of stack revTargetStack() to the bottom of stack tMenuBar + 20
end if
```

This patch makes sure that the correct handler (i.e. the one of stack `revMessageBoxBehavior`) will be called.

### DIFF
--- a/Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
+++ b/Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
@@ -90,19 +90,20 @@ on openStack
    
    lock messages
    set the minHeight of me to tMinHeight
-   moveStack
+   moveMsgBoxStack
    unlock messages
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok b63287a

livecode-vulcan:
💙 review by @livecodeali ok b63287a2e09f5d7e14bb1feebdfc08f6062505c2

PR #1162, opened 2016-05-06
Bugfix 17553

mwieder:
The tools palette in LC8 attempts to assign some very large numbers to the spray, brush, and eraser properties, and these numbers do not correspond to valid images. That causes an error which is not normally displayed to the users, since it happens in a system stack. This patch assigns the defaults listed in the dictionary to these tools on openStack, thus permitting these three tools to function properly.

Assigning brush=8, spray=34, eraser=2.


mwieder:
Nine commits? Am I submitting this against the wrong branch? I though develop-8.0 was for LC8 maintenance.

montegoulding:
Looks like you branched from develop

mwieder:
I pulled from livecode's develop-8.0, then created my bugfix branch from there. I haven't touched the develop branch in some time.

montegoulding:
Were you in the IDE submodule at the time? It could be the submodule pointers in the main repo aren't up to date.

mwieder:
Yep - in the ide submodule.

mwieder:
Have I mentioned lately how much I hate submodules?

montegoulding:
Well I'm not sure what's going on. I'm also not sure where you got the develop-9.0 branch from because I can't see it here. Might need to wait for @livecodeali to wake up and apply his genius. Alternatively you could rename your branch, checkout a clean branch of develop-8.0 named bugfix_17553, cherrypick c1090cb28e6d1329c26c9edfeb14de3a025891b3 and then force push to replace this PR with the cleaner version. You will also need to add a note.

mwieder:
That's "develop-8.0", and it's in the pulldown list of branches here. And yeah, I'll wait for Ali. No rush.

montegoulding:
Actually something else must be at play here @mwieder. The `reset paint` command should be doing what your patch does.

mwieder:
Granted, I'm probably patching over something that isn't working at a deeper level, but the patch does the trick. To see the problem, set gRevDevelopment to true, then try to select one of the three errant paint tools. Or bring up the message box and type 'put the eraser'.

mwieder:
Verified... the reset paint command sets the eraser to 4294967198.

montegoulding:
OK, probably better to close this then and try and work out why that's happening

mwieder:
peace out

montegoulding:
The issue appears to be attempting to set a negative value to an unsigned integer:

```
void MCInterfaceGetEraser(MCExecContext& ctxt, uinteger_t& r_value)
{
    r_value = MCeraser < PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
}
```

As you can see setting the property adds PI_BRUSHES correctly:

```
void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_value)
{
    uint4 t_newbrush = p_value;
    if (t_newbrush < PI_PATTERNS)
        t_newbrush += PI_BRUSHES;
```

So what we need to do is add PI_BRUSHES to all the defaults in reset paint:

```
void MCGraphicsExecResetPaint(MCExecContext& ctxt) 
{
    MCeditingimage = nil;

    MCbrush = 8+ PI_BRUSHES;
    ....
```

montegoulding:
@mwieder See my correction. Add `PI_BRUSHES` instead of `PI_PATTERNS`

livecodeali:
There was some jiggery pokery related to the fix for issue 1116 because I missed that it was targeted at develop and merged it, when it should have gone into develop-8.0 (see the top two PRs here)
https://github.com/livecode/livecode-ide/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+front

I'm not sure exactly what's going on but you might want to hard reset your local develop-8.0 to the upstream one and push it to your fork to get rid of the excess commits.

mwieder:
Ali - thanks, I have now hard reset the branch in the hopes that this will bring things back into alignment.

Regarding the fix to this, I think it might be more appropriate to do the adjustments in MCInterfaceSetBrush than in MCGraphicsExecResetPaint and have MCGraphicsExecResetPaint call MCInterfaceSetBrush to set the values.

PI_PATTERNS is 136
PI_BRUSHES is 100
So 
if (t_newbrush < PI_PATTERNS)
        t_newbrush += PI_BRUSHES;
still has a chance to be negative within a 36-value range.

Also, I see the default value of the spray brush is 31. The docs say 34. Are the docs wrong or should this be fixed in MCGraphicsExecResetPaint? 34 seems like a more appropriate spray value.

montegoulding:
@mweider the brushes are ID PI_BRUSHES (100) to PI_PATTERNS (136) so adding/subtracting PI_BRUSHES is correct and can't result in negatives. It might be more correct to check against PI_BRUSHES though as setting to 130 should just set as 130 instead of 230... there could be some legacy quirk that prevents that change though.

mwieder:
I dunno. That math looks fishy to me - it's an accident waiting to happen, as with the situation we're in now. Pass a small negative value for t_newbrush and you're in trouble. I agree that comparison with PI_BRUSHES would be more appropriate, and I'm willing to fix those two files the way I described above if I can get an ok from the team. Otherwise I'm not wasting more time on this.

montegoulding:
@mwieder I don't think you need to change MCInterfaceSetBrush at all if you are going to use it. It's probably a good idea though there is the side effect of throwing errors if the default brushes aren't available which is a new behavior and potentially problematic for some. Quite possibly not a bad behavior though seeing as it might be better to know that the IDs aren't available if you are intending to use them.

mwieder:
Well, if the default brushes aren't available there's probably a bigger problem afoot <g>. At any rate, an error is thrown now at the script level if the brush being assigned isn't available, which is what's causing the current problem.

montegoulding:
@mwieder I'm thinking in a standalone where the resources might not be there. The error isn't on `reset paint` it's on setting the brush to a non-existent image id. So that error would be thrown by `reset pain` tin a standalone if the standard brushes aren't there which would be a new thing. Perhaps it's correct behavior though... I must confess I don't think I've ever used the paint tools which I think makes me more reluctant to change their behavior because I may not be able to comprehend down stream consequences.

mwieder:
A couple of things here... first of all, I've got the modifications working locally here, and things seem to be ok. Secondly, I'm not sure the paint tools can be used in a standalone. There's been previous discussion of this on the list, and I don't remember the outcome. So it would probably good to have more team input into this before moving forward. Since I'm going to be out of town for a week I think I'll go ahead and submit a new PR with the changes in place and see what happens after that.

montegoulding:
good idea @mwieder & thanks!

mwieder:
Done. Submitted as https://github.com/livecode/livecode/pull/3996.

PR #1178, opened 2016-05-19
[Bug 16848] Force documentation HTML regeneration if missing

peter-b:
Sometimes the IDE documentation builder can get wedged -- it can think
that the documentation doesn't need regenerating, even though the
`api.html` file is missing from the documentation cache.

Sometimes, the documentation builder remains wedged even when the
whole `~/.runrev` directory is removed!

This patch changes the logic for ensuring the dictionary is available
to explicitly check for the presence (or otherwise) of the `api.html`
file.  Hopefully this will help unwedge some users dictionaries.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -10349,15 +10349,24 @@ private command revIDEGenerateDistributedAPI
 end revIDEGenerateDistributedAPI
 
 on revIDEEnsureDictionaryUrl pWhich
-   local tLastGenerated, tLastModified
+   local tNeedUpdate, tLastGenerated, tLastModified
    set the itemdelimiter to slash
    local tFile
    put revIDEGetDictionaryHTMLTemplate(pWhich) into tFile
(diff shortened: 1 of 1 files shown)


peter-b:
@livecodeali Does this look sane?

livecodeali:
Wouldn't it be better just to modify `revIDELastModifiedTimeOfFile` to return 0 in all the cases that it causes problems for - i.e. if `the result` after `revIDEPushDefaultFolder` is not empty or if the appropriate lineOffset is not empty? 

It is slightly odd because looking at the code it should be returning empty in all of these cases anyway. But if we could find a solution inside `revIDELastModifiedTimeOfFile` it would fix potential problems of this type in the (many) uses of the function elsewhere in the IDE library.

peter-b:
@livecodeali What about if I add a specific function for checking if file _x_ needs to be updated from file _y_? Then the problem of dealing with missing files etc. can be handled in one place and most of the places that currently use `revIDELastModifiedTimeOfFile` could probably be simplified a lot.

livecodeali:
V good idea!

livecodeali:
And could you add some tests for it?

PR #1191, opened 2016-05-26
set actual brush ids, not index

mwieder:
Submitting for consideration: this patch removes the legacy need to map brush ids 1..36 as indexes to the brush image ids 101..136, but rather sets the ids directly. All tests are passing and I have so far not uncovered any side effects. Note, though, that this requires a corresponding patch to the livecode repository, so I'm having to submit it separately.

Have I mentioned lately how much I hate submodules?


montegoulding:
@mwieder what about the scripts that set the brush?

mwieder:
reset paint is taken care of in the corresponding PR. This one takes care of the brush popup, which afaict is the only other place the brushes get set by id. The paint tools seem to function normally and the brush settings for the brush, eraser, and spray tools are correct.

mwieder:
I believe there are doc files that would need to be adjusted as well, now that you mention it.

montegoulding:
I'm talking about people's scripts

mwieder:
Not much chance of user scripts falling afoul of this:
1. the mapping algorithm has been broken since forever.
2. the mapping only applies to a small set of brushes (101..136) within the revicons stack; mapping to user images for brushes is unaffected.
3. setting the builtin brushes only applies within the IDE; using the paint tools in a standalone isn't supported and would take a lot of effort to implement, _especially_ because of the id mapping and the fact that the algorithm has always been broken.
4. there isn't any reason to set the built-in brushes by script within the IDE except as noted by the revBrushPopup stack.
5. I posted to the list to see if anyone might possibly have undertaken brush mapping, and there were no responses, as expected.

I'd like to remove the pattern mapping as well in a separate PR, and for the same reasons.

runrevmark:
Using the paint tools in standalones is supported and works fine as far as I know... Of course, you do need to ensure you include any images you want to use as brushes.

mwieder:
OK. I thought I had heard that the paint tools didn't work in standalones. Maybe they just haven't worked because of the brush mapping, although if you supply your own images that should be fine. With or without this PR.

montegoulding:
If someone clones the brushes stack and includes it in their app or has an old MC project that has it as a substack then I think this change would break their project. I'll try and have another look at the reset paint pr today to see what need to be done to merge.

mwieder:
Yes, and the other way this might break would be if someone deliberately sets the brush to a 1..36 value from a script in the process of doing something else. But do you seriously think that either of these scenarios are viable? (Except for me, and I've already adjusted PowerTools to take care of either situation). I stopped supporting MetaCard when I could count the number of users on one hand, and it's dropped to zero since then.

montegoulding:
I agree it's not a huge risk but we don't actually need to take the risk. After reviewing the bug reports on brushes and patterns there are people using them.

PR #1196, opened 2016-05-26
Add dependency checking primitive to IDE library

peter-b:
This changeset adds the `revIDEIsFilesetStale()` function to the IDE library.  This is a dependency-checking primitive for checking whether a set of derived files should be updated from a set of input files, and is intended to avoid the need to manually fetch and process file modification times.


livecodeali:
As far as I can get my head round this, it looks right!
@livecode-vulcan review ok 571a07f

livecode-vulcan:
💙 review by @livecodeali ok 571a07fb5f6fcbb34c31e12610b405eed1b7cca9

peter-b:
I want to do a bit more local poking at the documentation building to make sure it still works correctly after these changes, so I'm removing the milestone and I don't expect to include it in 8.1.0-dp-2.

peter-b:
Documentation still seems to work, as far as I can tell!

PR #1203, opened 2016-05-29
show  alpha value of gradients dynamically

BerndN:


Toolset/palettes/inspector/editors/com.livecode.pi.gradientramp.behavior.livecodescript
```diff
@@ -87,6 +87,7 @@ on updateColorValue pColor
    lock screen
    lock messages
    set the swatchColor of widget "color" of me to pColor
+   put pColor into field "value" of me
```
montegoulding: @BerndN This patch effectively makes `updateColorValue` and `updateColorDisplay` exactly the same. I have a tendency to think if there's two such similar implementations there must be a reason for the difference. I'm sure @livecodeali will know!

BerndN: @montegoulding updateColorDisplay is not called when changing the alpha value by slider but updateColorValue is. So I went with the conservative approach just adding the fix to updateColorValue without fiddling with messaging.
As you say, @livecodeali  will know.

livecodeali: Unfortunately I'm not quite sure why there are two such similar implementations! 

Originally I was using separate implementations for updating the visual feedback of the chosen color value and updating the stored editor value - it looks like in the course of fixing things these two handlers have become effectively identical, so I think there is no problem with merging them into one handler.

BerndN: @livecodeali 
I updated the pull request and propose now to take handler "updateColorValue" out and change in 
handler "updateSubeditorValue" to "updateColorDisplay"
As far as I can tell it works and has no untoward effects

livecodeali:
@BerndN Hi Bernd, your bugfix note's filename needs to be `bugfix-17759.md` and the description in it needs to start with a `#`.

BerndN:
@livecodeali Done

livecodeali:
@BerndN Thanks!
@livecode-vulcan review ok b45c5b8

livecode-vulcan:
💙 review by @livecodeali ok b45c5b80d11e617b60f65ddd8ebe2205462fc8f6

PR #1204, opened 2016-05-30
[[ Bug ]] Improve script only stack file handling of IDE save

montegoulding:
Previously the IDE would not check if a stackFile was script only and
therefore would determine that it was a legacy file format and not
provide warnings about data loss or present the user with the correct
file extension.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -7044,6 +7044,13 @@ function revIDEStackFileVersion pStackFile
       end if
 end revIDEStackFileVersion
 
+function revIDEStackFileIsScriptOnly pStackFile
+   open file pStackFile for text read
+   read from file pStackFile for 1 line
+   close file pStackFile
+   return matchText(it, "script \".*\"")
+end revIDEStackFileIsScriptOnly
+
 -- The revIDESaveStack and revIDESaveStackAs handlers use the "save"
 -- command with messages locked, which prevents the "saveStackRequest"
 -- message from being sent to the stack being operated on.  However,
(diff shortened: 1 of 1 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -7136,6 +7143,9 @@ on revIDESaveStack pStackID
    
    lock messages
    
+   local tStackFilename
+   put the effective filename of stack tStackName into tStackFilename
+   
    local tStackFileVersion
    
    if the cPreserveStackVersion of stack "revPreferences" is true and the cREVGeneral["stackfileversion"] of stack tStackName is not empty then
```
montegoulding: @livecodeali `cREVGeneral["stackfileversion"]` can be changed to `revIDEStackFileVersion()` I think. Or do we for some reason want the user to be able to save the stack with non-ide commands to a different format and then revert the stack when saving via the IDE command?

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -7044,6 +7044,13 @@ function revIDEStackFileVersion pStackFile
       end if
 end revIDEStackFileVersion
 
+function revIDEStackFileIsScriptOnly pStackFile
+   open file pStackFile for binary read
+   read from file pStackFile for 9
+   close file pStackFile
+   return it begins with format("%c%c%cscript",0xEF,0xBB,0xBF)
+end revIDEStackFileIsScriptOnly
```
peter-b, line 7052: In the future we might support script-only stacks without BOMs, so I suggest:
1. Open the file for binary read
2. Read 4 bytes from the file; if they are "REVO", return false
3. Re-open the file for text read (don't specify encoding)
4. Read 1 line from the file; if the line doesn't match `^script ".*"` return false
5. Return true

Toolset/libraries/revidelibrary.8.livecodescript
livecodeali: No, I think you're right actually - we can get rid of all occurrences of `cRevGeneral["stackfileversion"]`

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -7044,6 +7044,13 @@ function revIDEStackFileVersion pStackFile
       end if
 end revIDEStackFileVersion
 
+function revIDEStackFileIsScriptOnly pStackFile
+   open file pStackFile for text read
+   read from file pStackFile for 1 line
+   close file pStackFile
+   return matchText(it, "script \".*\"")
```
montegoulding, line 7051: @peter-b this is the best I can do until [17770](http://quality.livecode.com/show_bug.cgi?id=17770) is fixed because the `^` will find the BOM

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -7136,14 +7143,26 @@ on revIDESaveStack pStackID
...
    lock messages
    
+   local tStackFilename
+   put the effective filename of stack tStackName into tStackFilename
+   
    local tStackFileVersion
    
-   if the cPreserveStackVersion of stack "revPreferences" is true and the cREVGeneral["stackfileversion"] of stack tStackName is not empty then
-      try
-         put the cREVGeneral["stackfileversion"] of stack tStackName into tStackFileVersion
```
montegoulding, line 7143: @livecodeali I searched all stacks and could only find these two spots in the ide library with mention of `cREVGeneral["stackfileversion"]` so I think I've successfully killed it.

livecodeali, line 7143: 👍

peter-b:
@livecode-vulcan review ok e282f11

livecode-vulcan:
💙 review by @peter-b ok e282f1135ba9197282dfed1fb483944ebf7fd171

PR #1205, opened 2016-05-30
Set gRevDevelopment depending on if the engine is in the build dir

montegoulding:
Just an idea to see if people like it.

### DIFF
--- a/Toolset/home.livecodescript
+++ b/Toolset/home.livecodescript
@@ -449,6 +449,9 @@ private command revInternal__SetupPaths
    if tRepositoryPath is not empty then
       put tRepositoryPath into sRepositoryPath
       put tEnginePath into sBinariesPath
+      
+      global gRevDevelopment
+      put true into gRevDevelopment
    else
       put empty into sRepositoryPath
    end if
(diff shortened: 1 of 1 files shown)


livecodeali:
I don't see why not. 
@livecode-vulcan review ok d25d3ec

livecode-vulcan:
💙 review by @livecodeali ok d25d3ecc11ee8f4156befcef12c537fde74f8a82

mwieder:
Yes, please. I end up invoking the message box and doing this at every launch anyway.

livecodepanos:
I have been testing LC 8.0.2 rc1 (from the staging area) and it seems that now setting `grevdevelopment` to `true` has no result, since if I put a breakpoint in e.g. `revSaveAsAndroidStandalone` stack script, in `revSaveAsMobileStandaloneMain` command, the breakpoint is not respected. Moreover, typing `put grevdevelopment` in the msg box returns "grevdevelopment". So it looks like `grevdevelopment` is not treated as a global.

montegoulding:
Thanks @livecodepanos I'll check it out

PR #1208, opened 2016-05-31
[[ Bug ]] Ensure lowercase display name is used for associations check

livecodeali:
This happened to work with object types as the name of an object in
all cases where associations are present were identical to the lowercase
display name.


peter-b:
@livecode-vulcan review ok 7660036

livecode-vulcan:
💙 review by @peter-b ok 7660036c2c101a1b1fb1f9666f85b93645a5f128

PR #1216, opened 2016-06-04
[[ Bug ]] Ensure gREVDevelopment is init so it's accessible from msg box

montegoulding:


montegoulding:
@livecodepanos this should fix `gREVDevelopment` you should be able to do this in the msg box until this is merged:

```
global gREVDevelopment; put true into gREVDevelopment
```

livecodepanos:
Thanks @montegoulding. This patch sets `grevDevelopment` to `false` if it is empty, right? If this is the case, I was wondering if it would be clearer to make it:

```
global gREVDevelopment
if gREVDevelopment is empty then
  put false into gREVDevelopment
end if
```

Moreover, this patch should be submitted against `develop-8.0`

montegoulding:
@livecodepanos yes you are right it is clearer... it seems you can't [just edit](https://github.com/isaacs/github/issues/18) the PR to change the target branch when you accidentally forget that GitHub doesn't work out which branch you branched from unlike BitBucket... how annoying

PR #1211, opened 2016-06-01
[[ develop ]] livecode script guide updates

asayd:
- Removed reference to QT needed for Windows media playback.
- Line continuation character doesn't need to be escaped.
- Other minor formatting fixes.
- Changed references to "text boxes" to "note styles".
- Updated description of LiveCode documentation resources.
- Revised and updated description of Start Center.
- Links to new image resources.

PR #1228, opened 2016-06-08
[[ DontUseQT ]] Document QuickTime deprecation

montegoulding:


montegoulding:
Sorry @livecodeali I thought I was done with the IDE patches for QT but I found more!

livecodeali:
Weirdly the first part of this should be gone already (from here: https://github.com/livecode/livecode-ide/pull/1211/files)

livecodeali:
@livecode-vulcan review ok 0d4255b

livecode-vulcan:
💙 review by @livecodeali ok 0d4255bafde4299d8ddb593e50a1f65253277a50

montegoulding:
@livecodeali I'm not sure what happened there. I'll double check your other changes haven't been lost somehow...

montegoulding:
@livecodeali something odd has happened here as those changes are not in this file but the PR diff does not show them being reverted.

montegoulding:
@livecodeali It looks like I branched off develop-8.0 instead of develop for this PR, however, as far as I know the worst that should have caused is a merge conflict. I'm wondering if I managed to confuse git because the parent of this commit was merged into develop... I'm a bit disappointed in git if that's the case... I will check through the log to see if anything else is missing... 😞

montegoulding:
@livecodeali OK so it actually merged up all fine with nothing missing it appears... git must be ok with deletion from both branches and we managed to not have any whitespace diffs at that point. I have another PR coming with some minor patches to the use of the line continuation character which is used on the start of the second line rather than the end of the first.

livecodeali:
@montegoulding Ah that makes sense, and I guess with identical deletions it doesn't need to assign blame either.

PR #1236, opened 2016-06-13
[[ Preferences ]] Start documenting preferences

montegoulding:
A start on documenting preferences used by the IDE with the intention that these are moved to the API docs once complete and in the meantime added to as we work with various preferences.


peter-b:
@montegoulding This patch seems to include unrelated handler renaming changes.

montegoulding:
@peter-b Indeed it does... it has been a while since I did this but I vaguely recall recognising that I hadn't followed the private handler convention in this stack `__` I'll separate into a separate patch.

PR #1237, opened 2016-06-13
[[ Bug ]] Fix issue where interactive find was not resizing correctly

montegoulding:


peter-b:
Hi @montegoulding, I think this patch should go into the `develop-8.0` branch!

PR #1238, opened 2016-06-13
[[ Bug ]] Fix various multiple screen issues

montegoulding:
- Handle desktopChanged and ensure stacks are
  within the rect of their current screen
- Bug 17755 - Ensure modals on Mac aren't moved to
  the main screen by setting their location to the
  screenLoc
- Bug 17461 - Change revMenubar on Mac to modeless
  instead of palette so it doesn't cover fullscreen
  window widgets

Binary stacks that are included in standalones don't use the new `revIDEStackScreen*` APIs but make the same calculations themselves before setting their location on Mac. Some of these are actually highly unlikely to be used on Mac even though there is code handling Mac in them. They could be removed from this PR if desired.

I have set the milestone to 8.1.0-future mainly because of the change to revMenubar to be modeless. This is something I have had in a plugin for a long time without issues however it does allow stack windows to overlap the toolbar buttons so the change might need some live testing rather than just code review. It resolves the _stack maximized under toolbar buttons_ issue which is almost impossible to get out of without using the message box to set the top of the stack and the new even worse _stack fullscreen under toolbar buttons_ issue which is impossible to resolve because we have no script control to toggle fullscreen state.


livecodeali:
Personally I'd be happy with the change of reMenubar's mode too, so unless anyone else has any objections I reckon we should go with it. 
@livecode-vulcan review ok ad9e9fd

livecode-vulcan:
💙 review by @livecodeali ok ad9e9fdf0bf538bbea9283fa8d475149f9d06ba2

montegoulding:
I have another couple of minor tweaks I want to add to this so I'll temporarily
@livecode-vulcan review not ok ad9e9fd

I also have a more major change related to the window bounding rect which I think should go in another PR. Essentially it is centralising management of the windowBoundingRect and making it cover the entire virtual desktop rather than just the first screen. Ideally there would be a windowBoundingRects property but there isn't...

peter-b:
@montegoulding I edited your comment because instructions to @livecode-vulcan need to be at the start of a line.

montegoulding:
Ah.. thanks @peter-b

peter-b:
@montegoulding I've removed the "8.1.0-future" milestone from this... I assume it'll currently be targeted for 8.1.0-dp-3.

montegoulding:
Hmm... I think I set the 8.1.0-future milestone before an 8.1.0-dp-2 one existed on this repo. I think it could go in dp-2 unless it is too late for that.

Toolset/palettes/reverrordisplay.livecodescript
```diff
@@ -33,6 +41,9 @@ on preOpenStack
    local tStoredScreenRect
    put revIDEGetPreference("cErrorDisplayScreenRect") into tStoredScreenRect
    
+   local tScreenRect
+   put revIDEStackScreenRect(the short name of this stack) into tScreenRect
+   put tStoredScreenRect && tScreenRect
```
livecodeali: There's a rogue `put` here

livecodeali:
@livecode-vulcan review ok 035619c

livecode-vulcan:
💙 review by @livecodeali ok 035619c253f40f0bf6d857bc5b27972e08cde54a

PR #794, opened 2016-01-18
[Bug 15588] Script editor performance improvements

peter-b:
Improve script editor performance when working with huge scripts in a number of ways:
- Only reset the script editor field styles (font style / size) if they don't match the current preferences setting
- Align images in the script editor's handler list more efficiently (thanks to @runrevmark for the fix)
- Only resize script editor fields when necessary
- Cache the whole main script editor field to avoid re-populating when switching repeatedly between a small set of tabs

This is for @trevordevore :wink:

[Review on Reviewable](https://reviewable.io/reviews/livecode/livecode-ide/794)

PR #1239, opened 2016-06-15
[[ Bug 17146 ]] Fix issue where interactive find was not resizing correctly

montegoulding:


livecodeali:
@montegoulding I've put a breakpoint into `MCGroup::SetLockUpdates` and it doesn't look like it's ever being called - does this fix maybe work by virtue of the fact that `computeminrect(True)` is called when `lockUpdates` is set to false?

montegoulding:
Possibly... I could not think of any other property that would have caused the behavior we were seeing

livecodeali:
@montegoulding This patch doesn't actually work for me, strangely. So maybe not just a case of forcing computeminrect. 

Strangely, looking at where this was introduced (https://github.com/livecode/livecode-ide/pull/794/files), there doesn't appear to be any difference. However I tried replacing 

```
set the rect of tControls["find"] to tNewRect
```

with the ostensibly equivalent

```
      set the width of tControls["find"] to tMyRect[3] - tMyRect[1] - tGutterWidth
      set the left of tControls["find"] to tMyRect[1] + tGutterWidth
      set the top of tControls["find"] to tMyRect[4]  - tFindHeight
```

and it worked. So in short, I have absolutely no idea what's going on here.

montegoulding:
Hmm... Actually my patch doesn't work for me now! This is quirky to say the least.

I don't think we actually need to do anything but set the topLeft of the group and this appears to work fine: `set the topLeft of tControls["find"] to tMyRect[1] + tGutterWidth, tMyRect[4]  - tFindHeight`

If you have a chance to confirm that works for you I'll change my patch to use that, however, there's something deeper in this and I'd really like to know what it is. I can't think of any group property that would make a group behave this way!

livecodeali:
Will check it out but I suspect it will work.

I wonder if it's something to do with a `lockLoc` set to true on either the group itself or a containing group...

livecodeali:
@montegoulding Just tried it - setting the lockLoc of the owner of tControls["find"] to false, then setting the rect, then setting the lockLoc back to true works

livecodeali:
@montegoulding Your fix setting the `topLeft` also works for me.

montegoulding:
Hmm... looking at applyrect the following condition would likely be true:

```
rect.x + rect.width == nrect.x + nrect.width &&
rect.y + rect.height == nrect.y + nrect.height
```

Because what we are adding to the origin we are subtracting from the size. I'll check blame to see if this is a regression.

montegoulding:
@livecodeali That should fix it in this case, however, I'll follow up to determine what to do about the group resizing quirk. If its by design it should be documented somewhere... I can't work out what the design would target though... git log shows it has been this way since at least LC 6.

livecodeali:
@livecode-vulcan review ok 24379fe

livecode-vulcan:
💙 review by @livecodeali ok 24379fe91c48a807ef9d26b0b6fcc72bf8550e7b

PR #1242, opened 2016-06-18
[[Community Docs]] updated headers in LiveCode Script guide

jameshale:
Headers in script guide were too deep for sections to appears in the Guide window.
Dropped most header levels down one "#" to enable them to show in lefthand panel of guide window.


jameshale:
OK so this didn't work as expected.
Perhaps instruction need modification for Guide updates.

PR #1244, opened 2016-06-23
[[ ExternalDependencies ]] Support LCS based dependencies for externals

montegoulding:
This patch provides support for externals found
in the Tools/Ext directory to include supporting
LCS based stackfile dependencies and for those
dependencies to manage dependencies between
themselves and other standalone inclusions.

In the course of development some unnecessary code
was removed from revDeployLibrary and revSBLibrary
was elevated to an always loaded IDE back script.
Additionally the recursive stackfile lister was
refactored.

There are dependencies between this patch and an
associated patch to the livecode repo.


Toolset/home.livecodescript
```diff
@@ -1207,6 +1208,28 @@ command revInternal__InitialiseLibraries
...
+         set the folder to revEnvironmentExtPath() & "/" & tExternal
+         
+         local tFiles
+         put the files into tFiles
+         filter tFiles with "*lib*.livecode*"
+         
+         repeat for each line tFile in tFiles
+            if there is a stack tFile then
+               start using stack tFile
+            end if
```
techstrategies, line 1231: @montegoulding Just thinking through how the external libraries work here.... Rather than issuing a "start using", would it be better to do:

`send "revLoadLibrary" to stack tFile`

so that the script can choose to insert itself into the back, set itself as a behavior, or whatever it wants to do?

montegoulding, line 1231: Yes that's one of the things I need to change along with the initialisation library that should remove a stack from memory if it doesn't handle revLoadLibrary

montegoulding:
Leaving as WIP while I sort out branches with the correct submodule pointers so I can test this

livecodeali:
@livecode-vulcan review ok 6d552a5

livecode-vulcan:
💙 review by @livecodeali ok 6d552a5f377eb262705270d4bd03af88106afd0a

PR #1258, opened 2016-06-29
[[ IDE ]] Include info about file that causes dependency check error

livecodeali:
Previously if an error was thrown during revIDEIsFilesetStale, the
filename causing the error was not included in the error message.


livecodefraser:
@livecode-vulcan review okay 457de3e

livecode-vulcan:
💙 review by @livecodefraser ok 457de3e4bac54ada12b0be72be09e45dafda0955

PR #1268, opened 2016-07-02
Check the target is me in revLoadLibrary and revUnloadLibrary

montegoulding:
There is no specific bug related for this but it does reduce the potential for:
- setup code in `revLoadLibrary` to be run multiple times if some back script didn't load correctly
- teardown code to be run on the wrong library in the event a library was not loaded correctly and then `revUnloadLibrary` was sent to it.

### DIFF
--- a/Toolset/libraries/revanimationlibrary.livecodescript
+++ b/Toolset/libraries/revanimationlibrary.livecodescript
@@ -12,240 +12,289 @@ end revUnloadLibrary
 local lOverRide
 
 on revGoToFramePaused pWhichAnimation, pWhich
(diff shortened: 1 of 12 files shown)


montegoulding:
Additionally because explicitVariables is enforced to true now there were a few older libraries requiring some variable declarations. Where I had to do significant changes I deliberately reformatted handlers with the enforced indent.

livecodeali:
@livecode-vulcan review ok 21bdd05

livecode-vulcan:
💙 review by @livecodeali ok 21bdd0548e3d007852daf5055e118f230afc781a

PR #1265, opened 2016-07-01
Force some prefs while gREVDevelopment is true

montegoulding:
- script indent = 3
- explicitVariables = true
- preserveVariables = true
- comment char `--`
- gREVShowStacks = true

PR #1269, opened 2016-07-02
Don't set IDE scripts that don't compile

montegoulding:
Previously there was a check to ensure script editor
scripts were not set if they didn't compile. This has
been extended to all scripts when gREVDevelopment is
true.

This change makes it **much** simpler to edit
IDE scripts, however, the team needs to be aware that
scripts with compile errors will _not_ save. If necessary
set gREVDevelopment to false temporarily to work around.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
@@ -137,7 +137,9 @@ end revSEAddTargetObject
 #   Removes the specified object from this script editor. If the removed object is the current
 #   editing object, then changes the current object to the last selected object.
 command revSERemoveTargetObject pObject, pDontWarn, pDontCloseIfLast, pDontChangeTab
-   if pDontWarn is not true and seGetObjectState(pObject) is "edited" then
+   local tState
+   put seGetObjectState(pObject) into tState
+   if pDontWarn is not true and (tState is "edited" or (tState is "error" and revIDEObjectIsOnIDEStack(pObject))) then
       revSEWarnBeforeClosing pObject
       if the result is "cancel" then
          exit revSERemoveTargetObject
@@ -743,7 +745,9 @@ command revSEGetModifiedObjects
    
    local tEditedObjects
    repeat for each line tObject in tObjects
-      if seGetObjectState(tObject) is "edited" then
+      local tState
+      put seGetObjectState(tObject) into tState
+      if tState is "edited" or (tState is "error" and revIDEObjectIsOnIDEStack(tObject)) then
          # OK-2009-01-17 : Bug 7169
          local tDirty
          send "getDirty tObject" to group "Editor" of me
@@ -1479,7 +1483,7 @@ end seResizeCardBottom
 
 # Parameters
 #   pObject : the object to apply the script of
-#   pIgnoreErrors : causes errors in setting the script not to be displayed in the errors pane
+#   pIgnoreErrors : causes errors in setting the script not to be displayed in the errors pane and IDE scripts to be set regardless of errors
 #   @rCompilationErrors : gets the list of compilation errors placed into it. Will be set to empty if none occurred.
 # Returns
 #   True if the script was applied, false otherwise. Note that compilation errors may occur but the script could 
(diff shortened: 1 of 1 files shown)


mwieder:
What will happen if users set this to true and then edit scripts? Currently non-compiling scripts can be saved as works in progress, and this is useful when editing someone else's long scripts. Will there be an indication that the script can not be saved or will this just be a silent failure? The latter would be a red flag for me.

montegoulding:
At the moment with this patch there will be a silent failure to set the actual object script although the script editor will highlight the errors so it will be rather obvious as long as people know that `gREVDevelopment` causes this behavior. This has been the case for the script editor scripts for nearly 5 years. It's worth noting that setting `gREVDevelopment` to `true` is only really intended for people working on the IDE. Note that there is another [patch](https://github.com/livecode/livecode-ide/pull/1265) that forces some preferences based on `gREVDevelopment`.

mwieder:
Hmmm... maybe I've used glx2 for so long that I've forgotten how the IDE's editor works, but with glx2 it's certainly possible to save works in progress even if they don't compile. Ah! can you still save the script but not compile it? I'd be fine with that.

montegoulding:
I'm not sure what you mean there. The entire point is we _dont_ want to set the script of the object to a script that doesn't compile because doing so removes all its handlers from the message path and this causes _uncompiled critical IDE script hell_ syndrome

mwieder:
heh... yes, I know that syndrome. But saving the text is different from setting the script. And that's the way glx2 has always worked.

mwieder:
...and I see you put a notice in the use-list, so I'll move commenting there.

montegoulding:
OK, I don't know how glx2 works but the IDE has an intermediary object it uses to compile the script to check for errors and then it applies the script to the object itself. What this does is skip that last step if there's a compile error and gREVDevelopment is true. If we were implementing this for general use we would most definitely want some warnings at various points.

mwieder:
Yes, I think that's the only way to do this (compile, check for errors, set if possible). As long as there are annoying warnings and it's still possible to save the script text without setting it then I'm happy. I wouldn't want to spend a couple of hours working on a script, still not have it compile-ready, and not be able to save what I had done before going to bed. Working on someone else's 15k-line script (I've been there) gets to be an all-or-nothing marathon in that case.

But saving the script/stack and setting the script are two different things to talk about, so I think we're ok.

runrevmark:
This is certainly a neat idea for when you are editing certain IDE scripts . However, I tend to run with gREVDevelopment true when I'm _not_ editing the IDE just so I can see if there are any IDE errors being thrown which I can report to get fixed. Also, with the patch which sets gREVDevelopment true (which I believe was recently merged) when building from source, there will be even more scenarios when (even us) will be running and using the IDE when we aren't actually editing it (i.e. pretty much me, all the time when doing engine dev).

I wonder if this extra 'bit of safety' should be tied to whether a stack is an IDE stack, rather than all stacks. It should also probably 'do something' in the script editor window when a script didn't compile and thus has not been set on an object so that you can visually see that if you close the script editor, you might lose work.

mwieder:
Yes, it's the "losing work" part that troubles me. I also run with gRevDevelopment true much of the time, partly because I'm working at a system level quite a bit, and also partly because it's enabled me to catch and submit fixes for some IDE errors that have previously gone unnoticed.

montegoulding:
OK, I've tweaked this a little so that:
- the new behavior is _only_ for IDE scripts
- the same dialog that appears when closing an unapplied tab or the editor with unapplied scripts is presented
- in the event that the user wants to save a work in progress on IDE scripts (there _are_ many that won't break the IDE) then they can force apply from here due to `pIgnoreErrors` being true
- because the status of the script is `error` rather than `applied` the icon on the tab is (and was in the previous commit) red and errors are displayed so it is clear that the script is not applied (green) or modified unapplied (yellow).

I actually think this new IDE behavior would work well in all cases. The only significant hurdle would be crashes while sorting out compile errors but we would like to avoid them regardless...

montegoulding:
The script editor save handling probably needs some sorting out as at the moment it will appear to save but the erroring script won’t be applied. I’m looking into that now. I think in the event the script application couldn’t happen it should beep at you or something rather than save.

montegoulding:
Hmm... Actually the save behavior is a bit odd. It applies all scripts but only saves the stack with the currently open tab. What if you didn't want some other edit saved? I guess it is a helper if you are eating multiple scripts from the same stack though. Perhaps it should save all stacks too though as it does give that appearance?

montegoulding:
Actually I misread the code and it only applies scripts from the stack of the currently being edited script which makes much more sense!

mwieder:
Thanks, @montegoulding this is sounding good to me now.

livecodeali:
@livecode-vulcan review ok c238c51

livecode-vulcan:
💙 review by @livecodeali ok c238c516c55e84ad26294b077a9966e19312fc53

montegoulding:
Thanks @livecodeali once we have used this for a bit to confirm it covers all bases we can look at enabling this for everyone I think.

PR #1274, opened 2016-07-09
[[ Bug ]] ensure iconGravity is empty when creating buttons in IDE

montegoulding:
Previously the iconGravity was not being copied
during button creation which had the effect of
iconGravity always being empty on new buttons.

When generating the UI for revTools and revMenubar
the iconGravity of the templateButton was being
set to top but as the iconGravity from the
templateButton wasn't being copied by the engine
the UIs were designed with iconGravity empty.

This patch also ensures the templateButton is reset
before generating revMenubar

### DIFF
--- a/Toolset/palettes/menubar/revmenubar.livecodescript
+++ b/Toolset/palettes/menubar/revmenubar.livecodescript
@@ -321,6 +321,8 @@ on generateMenubarUI
       delete group "toolbar" of me
    end if
    
+   reset the templatebutton
+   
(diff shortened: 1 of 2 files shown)


peter-b:
@montegoulding Thanks so much for fixing this!

@livecode-vulcan review ok 1448750

livecode-vulcan:
💙 review by @peter-b ok 144875097a0b5547215de3a71c333006b5aaf5fd

PR #1276, opened 2016-07-09
[[ Bug 17718 ]] Ensure property inspector is visible

montegoulding:


peter-b:
@livecode-vulcan review ok 17718

@montegoulding It would be great if the commit message explained what this change fixes. If in the mode where IDE UI elements are hidden while editing the script editor, is it necessarily correct to show the PI? Maybe the "target object" of the PI should be reset but it should remain hidden until the palettes are re-shown again (or something).

montegoulding:
Hmm another possible fix would be to show the palettes when using the contextual menu as that's how you can get into this situation at the moment...

montegoulding:
Thinking about this I think if the user takes some action to present an inspector while palettes are hidden then it's appropriate that just the inspector is shown. It definitely should be shown regardless of how we do it.

montegoulding:
@peter-b just an improvement of the commit message here

peter-b:
@livecode-vulcan review ok dfe8136

livecode-vulcan:
💙 review by @peter-b ok dfe81367aaf57c4c313d4b48ac054b7e6265fb1e

PR #1277, opened 2016-07-09
[[ Bug 7157 ]] Ensure _ideoverride property of IDE stacks is set to true

montegoulding:


Toolset/libraries/revfrontscriptlibrary.livecodescript
```diff
@@ -287,20 +289,6 @@ on openStack
...
-   local tTarget
-   put the long id of the target into tTarget
-   if tTarget is not the long id of me then
-      local tStack
-      put revTargetStack(tTarget) into tStack
-      if revDebuggerEnabled() then
-         revDebuggerActivateBreakpoints the long id of stack tStack
-      end if
-   end if
-
```
peter-b, line 303: @montegoulding These changes don't seem to be related to the `_ideoverride` propery, as far as I can tell?

montegoulding, line 303: Ah, I'll split the commit. It was just some redundant code I found because it's now in preOpenStack

peter-b:
@livecode-vulcan review ok ff95e7e

livecode-vulcan:
💙 review by @peter-b ok ff95e7e29d2ef1cf8c2d198010296ed61f34afa6

runrevmark:
@montegoulding: Just checking, but does the _ideOverride thing work for script only stacks? I think that custom prop is checked by the engine on load... Which would happen before the IDE script has a chance to set it.

runrevmark:
(Also, I think the property name is 'ideOverride', not '_ideOverride')

@livecode-vulcan review not ok ff95e7e

livecode-vulcan:
💔 review by @runrevmark not ok ff95e7e29d2ef1cf8c2d198010296ed61f34afa6

livecodeali:
@runrevmark There is a custom property `ideOverride` and a development mode engine property `_ideoverride` - if the custom property is set when the stack loads then the engine sets the `_ideoverride` engine prop

montegoulding:
@runrevmark @livecodeali yes there is both ideOverride custom property and _ideoverride engine prop. I presume at some point someone didn't want to change the file format to add _ideoverride as a saveable property.

runrevmark:
Ah! Yes - that makes sense... That person would be me - I had just completely forgotten how it worked!

@livecode-vulcan review ok ff95e7e

livecode-vulcan:
💙 review by @runrevmark ok ff95e7e29d2ef1cf8c2d198010296ed61f34afa6

runrevmark:
Actually I wonder if the ide stacks should set the _ideoverride of themselves to true in preOpen or such (or in Home) and the ideOverride customprop logic removed (although, this thought does not mean this patch should not be included!).

montegoulding:
@runrevmark I'm fairly sure (not 100% positive as I'm still feeling my way around the IDE) that this patch will mean that the `ideOverride` custom property logic could be removed as all the IDE stacks should be having `_ideoverride` set anyway. There may be some exceptions with things that are opened in a `lock messages` environment.

PR #1278, opened 2016-07-09
[[ Bug 17409 ]] List all loaded stacks in the behavior picker

montegoulding:
Previously only open stacks were listed in the
property inspector behavior picker. As many stacks
are loaded into memory for use as behaviors and
libraries but not opened it is more helpful to
list the mainstacks and their substacks.


peter-b:
@livecode-vulcan review ok b9f30a7

livecode-vulcan:
💙 review by @peter-b ok b9f30a7eec9d0f22dacb19bb5d7d8bc305de0718

PR #1283, opened 2016-07-13
[[ Bug 11834 ]] Remove deprecated syntax from iOS card of standalone settings

livecodeali:
It is not needed and causes information loss


peter-b:
@livecode-vulcan review ok e3bb50f

livecode-vulcan:
💙 review by @peter-b ok e3bb50f03e223e931eb3f730dfff710c531d974f

PR #1289, opened 2016-07-22
[[ Bug 17600 ]] Fix data grid refresh button in PI

livecodeali:
Replace the com.livecode.pi.datagrid PI editor (which
contained buttons to trigger refreshing the data grid and opening
the row template stack) with a generic com.livecode.pi.action editor
consisting of a single button, which forces an update whenever it is
clicked.

Call a custom setter when the buttons for these two data grid actions
are pressed, to handle the resulting action.


peter-b:
@livecodeali The IDE library feels like the "wrong" place for the `revIDEDataGridAction` handler to live, but I don't really have a better suggestion at the moment.

@livecode-vulcan review ok d50c0c9

livecode-vulcan:
💙 review by @peter-b ok d50c0c986c8bfae156f9b96194aa354f7dfe0e0f

livecodeali:
@peter-b All the data-grid support stuff should probably be in a separate backscript I guess. But since the `revIDEGet/SetDataGridProperty` functions are in there, it's the best place for the moment.

PR #1304, opened 2016-08-05
[[ Tutorial ]] Allow display of browser with given url in tutorial stack

livecodeali:


montegoulding:
Hmm... I was about to review ok when I noticed the feature note uses a video example. Wouldn't it be better to use a player if playing a video is the goal? That way the step progress/completion can be measured reasonably well using player events.

livecodeali:
I think video would be a separate (and probably sensible) enhancement, but this was what was requested. I don't think the intention was to have the video be a full tutorial, just to illustrate one step, although it would be pretty cool to have player callbacks trigger progression of the tutorial!

montegoulding:
@livecodeali it might be nice to have two options. One for a long running video that can play alongside the tutorial and has callbacks tied into steps and one for video as a step. Either way:

@livecode-vulcan review ok 26f7d0e

livecode-vulcan:
💙 review by @montegoulding ok 26f7d0e9fed53730ce3d395d4a2498b999ef5a05

tfabacher:
Thanks Ali...we are now in the process of getting this assigned. We will have one full time person creating lessons. We are just going to display the video in a browser, so it should not be much of an issue.

--Todd

PR #1307, opened 2016-08-09
[[ ColorizeMsgBox ]] Apply script editor colorization to msg box

montegoulding:
This patch fixes an issue with the message box
where pasting styled text was not being handled
appropriately. It also uses script colorisation
and user script background color preferences to
give the same look to message box scripts as script
editor scripts.

A small change to `revpreferencesgui` was required
to apply script editor preferences using
`revIDESetPreference` so we can subscribe to changes.
A future patch should use `idePropertyChanged`
subscription to update the script editor.


peter-b:
Hi Monte, it looks like this fixes [bug 18134](http://quality.livecode.com/show_bug.cgi?id=18134), so could you please add a bugfix note?

montegoulding:
@peter-b sorry I didn't know there was a bug reported

peter-b:
@montegoulding No problem, that's what @livecodepanos and I are here for.

peter-b:
@montegoulding I wonder if there should be a "ideColorizeField" command for syntax-highlighting a whole field?  It would save copy-and-pasting the whole `char 1 to the number of chars in...` dance

If you think it's probably fine without a helper command go ahead and mark this as reviewed by me.

montegoulding:
@peter-b because I can't think of another use case for a colorized field and the script editor I think sometimes only colors substrings I think I'll leave this. If we come up with other use cases it's an easy enough change.

@livecode-vulcan review by @peter-b ok aae49fe

livecode-vulcan:
💙 review by @peter-b ok aae49fe148ab2499111142d015913064ff271bc6

PR #1310, opened 2016-08-10
[[ Bug 18053 ]] Use appropriate cursors for selection handles

montegoulding:
This patch makes the following changes:
- Use appriorate resize cursors for each
  selection handle depending on the direction of drag.
- Remove the restriction for message subscribers to
  be open stacks (mode 0) as this was unnecessarily
  restrictive to backScripts etc.
- Add the publishing of ideMouseMove to the message
  handler library


Toolset/libraries/revbackscriptlibrary.livecodescript
```diff
@@ -3561,3 +3563,65 @@ private function utilityPathIsCaseSensitive pPath
   
   return tIsCaseSensitive
 end utilityPathIsCaseSensitive
+
+constant kHandleSize = 5
+
+on ideMouseMove pX, pY, pTarget
```
peter-b, line 3576: @montegoulding Could you please add some comments here that explain what this handler is for?

peter-b:
@livecode-vulcan review ok f1c397f

livecode-vulcan:
💙 review by @peter-b ok f1c397f3ddd4b6daa3d379e859d26c2789a46104

PR #1311, opened 2016-08-11
[[ Bug 18082 ]] Improve script only stack opening

montegoulding:
This patch makes the following changes:
- Script only stacks will now be checked to determine if they are visible
  after opening before editing their script in the script editor after
  opening via the file menu and when opened via the system. This allows
  script only stacks that generate their own user interface via script
  in `preOpenStack` or `openStack` to not open into the script editor
  by setting the visible of the stack to true in `openStack`.
- StackFiles are now explicitly opened at startup by the IDE rather
  than depending on the engine to open them.
- StackFiles opened via AppleEvent will now use the same code path
  as StackFiles opened via command line arguments and relaunch arguments

### DIFF
--- a/Toolset/home.livecodescript
+++ b/Toolset/home.livecodescript
@@ -288,7 +288,16 @@ on revInternal__openStack
       
       -- Now the home stack has loaded, we can allow interrupts..
       set the allowInterrupts to true
+      
+      revInternal__OpenCommandLineFiles
 end revInternal__openStack
+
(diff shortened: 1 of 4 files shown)


montegoulding:
Depends on [this PR](https://github.com/livecode/livecode/pull/4348) as otherwise the stacks will be loaded twice

montegoulding:
Marking as WIP until the dependant PR is reviewed

PR #1316, opened 2016-08-11
[Bug 18166] Script Editor menubar "Window" menu must show names of open stacks

livecodepanos:


peter-b:
@livecodepanos Would it be appropriate to simplify this by using "if the stack is an IDE stack then show the title, otherwise show the name"?  There is already a function for checking if a stack is an IDE stack in revidelibrary

livecodepanos:
In fact we do not want _any_ IDE stack to appear in the Window list. Only user stacks, script editors, the dictionary and resource center are considered windows from the point of view of the menubar. But indeed this can be simplified, since there is also a function that checks exactly this (`revIDEStackIsWindow`) in revidelibrary. I'll update the PR

peter-b:
Yay! Thanks @livecodepanos

livecodepanos:
Hmm in fact `revIDEStackIsWindow` does not filter out users stacks. However, the IDE stacks that we do not want to appear under the "Window" menu are already filtered out with `revIDEUpdateWindowList`. So we are safe to check just if the stack is an IDE stack :)

peter-b:
@livecode-vulcan review ok b12be22

livecode-vulcan:
💙 review by @peter-b ok b12be22fafc51d6e7c1d98f537baa263a595b8f4

PR #1320, opened 2016-08-12
[Bug 18174] Script Editor handler list sorting improvement

livecodepanos:
Special thanks to @Rinzwind for the suggested fix

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsehandlerlistbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsehandlerlistbehavior.livecodescript
@@ -145,7 +145,7 @@ command update
       # settings we sort alphabetically anyway, it probably doesn't make much difference to just sort here.
       # This will produce an ordering F, G, H, PF, PH, S, which may need to be adjusted using a custom search
       # function here.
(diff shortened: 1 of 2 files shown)


peter-b:
@livecode-vulcan review ok 7492540

livecode-vulcan:
💙 review by @peter-b ok 7492540586c51adbc39a458e34f1091290db4e3f

Toolset/palettes/script editor/behaviors/revsehandlerlistbehavior.livecodescript
```diff
@@ -145,7 +145,7 @@ command update
       # settings we sort alphabetically anyway, it probably doesn't make much difference to just sort here.
       # This will produce an ordering F, G, H, PF, PH, S, which may need to be adjusted using a custom search
       # function here.
-      sort tHandlers numeric by handlerTypeSortCallback(word 1 of each)
+      sort tHandlers by handlerTypeSortCallback(word 1 of each) & word 2 of each
```
montegoulding, line 148: hmm... should we add a note somewhere that if we ever end up with more than 9 handler types we need to switch to A...Z instead of 1...N. Alternatively this could be split into two sort commands by sorting in handler name order first.

PR #1321, opened 2016-08-12
Fix revIDEGroupObjects dependency on object selection

montegoulding:
This is a PR I had against develop-8.0 that I have now refactored to take advantage of the new `return` and `group` command features in 8.1. It also adds some user feedback (beep) where appropriate if grouping was not possible (most likely when selecting some grouped objects and some ungrouped objects and trying to group them).

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -3548,37 +3548,66 @@ on revIDEActionGroupObjects
    
    revIDEGroupObjects tSelobj
    
-   local tNewGroup
-   put the result into tNewGroup
-   if exists(tNewGroup) then
-      select tNewGroup
+   if the result is empty and exists(it) then
+      select it
+   else
+      beep
    end if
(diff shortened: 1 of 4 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -3548,37 +3548,67 @@ on revIDEActionGroupObjects
...
+   local tObject
+   repeat for each line tObject in pObjectIDs
+      if not exists(tObject) then
+         return "Object does not exist" for error
+      end if
+   end repeat
    
-   put the long id of (line 1 of pObjectIDs) into tObj
-   put 1000000 into tLayer
+   put line 1 of pObjectIDs into tObject
```
livecodeali: We don't need to do this anymore do we?

livecodeali:
@livecode-vulcan review ok e250221

livecode-vulcan:
💙 review by @livecodeali ok e2502214ca3022015fb0389dd99022369905f03b

PR #1322, opened 2016-08-15
[[ FilterHandlers ]] Clear filter when viewing a different script

montegoulding:
Previously the handler filter would remain when switching scripts and
cause confustion when no handlers were listed. This patch clears the
handler filter each time the user views a different script. This
patch also ensures that the script editor isn't unnecessarily
updated when the user clicks on the active script tab.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
@@ -41,6 +41,12 @@ local sLastShownExecutionPoint
 #   Sets the current object to edit to pObject. If pStartingLine is given, then seeks to that line
 #   otherwise seeks to the last selected line.
 command revSESetCurrentObject pObject, pStartingLine, pDontSendCallbacks
+   lock screen
+   
+   revSEGetCurrentObject
+   if the result is not empty and the long id of pObject is the long id of the result then
+       exit revSESetCurrentObject
+   end if
    # Check that pObject is one of the target objects, if not then just do nothing
    local tTargetObjects
    revSEGetTargetObjects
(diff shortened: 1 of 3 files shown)


montegoulding:
This does not appear to be in the bug db

livecodeali:
@montegoulding Do you think the logic of 

```
if tClickedTabNumber is not sSelectedTab then
```

should actually go into `revSESetCurrentObject` (i.e. don't do anything there if the object's tab is currently selected)? Not sure if that's possible. But I'm slightly concerned that various other places that call `revSESetCurrentObject` will cause the filter to be cleared even if it is the current tab.

montegoulding:
@livecodeali yes that probably makes more sense

Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
```diff
@@ -41,6 +41,12 @@ local sLastShownExecutionPoint
 #   Sets the current object to edit to pObject. If pStartingLine is given, then seeks to that line
 #   otherwise seeks to the last selected line.
 command revSESetCurrentObject pObject, pStartingLine, pDontSendCallbacks
+   lock screen
+   
+   revSEGetCurrentObject
+   if the result is not empty and the long id of pObject is the long id of the result then
```
montegoulding, line 47: @livecodeali we really need an equality operator for object references... it's a bit frustrating to need to do stuff like this to ensure you are comparing apples with apples.

livecodeali, line 47: @montegoulding Hmm, yeah... perhaps we should implement a `revIDEObjectIsObject` function which could be simplified if an engine function is added?

montegoulding, line 47: That's a good plan. @peter-b could we add that to the backlog? Maybe it could be part of the object handle variable type we discussed? FYI @livecodeali it was a value with an underlying pre-resolved object reference. When used as a string it would have a long id representation but otherwise need no parsing as it was used.

peter-b, line 47: @montegoulding Add a Bugzilla ticket please.

montegoulding, line 47: http://quality.livecode.com/show_bug.cgi?id=18188

livecodeali, line 47: Sounds good to me!

livecodeali:
@livecode-vulcan review ok 7a43ee3
Regardless of the object comparison issue, this looks good.

livecode-vulcan:
💙 review by @livecodeali ok 7a43ee36c1b5af2546a343d04d5e97dc79e3c718

PR #1328, opened 2016-08-16
[[ Bugs 17863 ]] Message box autocompletion issues

montegoulding:


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -10931,7 +10935,7 @@ private command __tryToDoScriptInObject pObject, pToDo, @rValidScript
    put the result into tError
    if tError is empty then
       try
-         send "-- execute in object context" & return & tDoScript to pObject
+         send tDoScript to pObject
```
livecodeali: The extra bit of comment here is to make sure that the engine uses `domess` instead of calling a handler in the target object directly - without it, I believe if you do something like:

```
myHandler "Param1", "Param2"
myOtherHandler "Param1", "Param2"
```

this will cause `myHandler` to be called with three parameters:
1. "Param1"
2.  "Param2"
myOtherHandler "Param1"
3. "Param2"

livecodeali:
@montegoulding Could you add some tests for these please? There is a bunch of similar stuff already in `ide/tests/messagebox/execution.livecodescript`

Toolset/libraries/revidelibrary.8.livecodescript
montegoulding: Hmm... it sounds like we want a way to call domess directly without trying to pre-parse parameters

livecodeali: Yes... yet another synonym of dispatch/post/message/send/call etc

montegoulding: `do <script> in object`?

montegoulding: However, it would be different from `in caller` as there's no context to use...

livecodeali:
@livecode-vulcan review ok 2973811

livecode-vulcan:
💙 review by @livecodeali ok 2973811130584be1fc9a3dde209d5cbc20b9646d

PR #1335, opened 2016-08-17
[[ Bug 17660 ]] Ensure ideSelectedObjectChanged message is sent by the IDE

montegoulding:


livecodeali:
Under the current system of the IDE, this should actually `lock messages` around the `do "select..` and then `ideMessageSend "ideSelectedObjectChanged"` after unlocking. (see note here http://quality.livecode.com/show_bug.cgi?id=18076)

However, maybe it's time to change that and remove all such lock messages in the IDE library.

montegoulding:
Hmm... I see no real reason to deliberately lock messages here. I'm not overly sold on the idea that anything the IDE does to user stacks will be done in a lock messages environment although it does avoid the every plugin has a frontscript

montegoulding:
@livecodeali I will patch this and we can cross this bridge later

livecodeali:
I think you're right, but I also think for consistency this should do what everything else does for now, until we change everything.

montegoulding:
I notice there's lots of `lock messages` then `unlock messages` in the code near here also which probably needs to to what I've done here and store the `lockMessages` for resetting.

montegoulding:
http://quality.livecode.com/show_bug.cgi?id=18218

montegoulding:
@livecodeali I have updated the note and commit summary to reflect the changes to the patch

livecodeali:
@livecode-vulcan review ok fec463c

livecode-vulcan:
💙 review by @livecodeali ok fec463c248532d0af8c73288dc994818cda3ce00

PR #1338, opened 2016-08-17
[[ Bug 14217 ]] Fix PB keybaord nav in edit mode

montegoulding:


livecodeali:
@livecode-vulcan review ok 23b814c

livecodeali:
We really need some helper functions that clarify what on earth things like

```
(the selectedField is empty and (the selObj) is empty) or (the selectedField is not empty and revTargetStack(the long id of the selectedField) is not the short name of the topStack))
```

are supposed to actually be checking, don't we? :-)

livecode-vulcan:
💙 review by @livecodeali ok 23b814c3b05ea0173f1e9a026dbbdd92d35d2a48

montegoulding:
@livecodepanos I just realised I forgot to investigate the return key issue and will do that today in a separate PR

PR #1347, opened 2016-08-19
[[ Bugs 17693, 18047 ]] Contextual menu issues

montegoulding:
This is a mini-moomin on the use of contextual menus. We need to do more at some point to enable some of the menus that currently require selectability such as cut, copy & paste operations so we can do those on closed stacks/cards etc

### DIFF
--- a/Toolset/palettes/inspector/revinspector.livecodescript
+++ b/Toolset/palettes/inspector/revinspector.livecodescript
@@ -99,7 +99,7 @@ on ideInspectObjects pObjects
    palette stack tTargetInspector
    unlock messages
    set the cSelectedObjects of stack tTargetInspector to tSelectedObjects
-   set the cLocked of stack tTargetInspector to tSelectedObjects is not the selectedObjects
+   set the cLocked of stack tTargetInspector to tSelectedObjects is not revIDESelectedObjects()
    unlock screen
 end ideInspectObjects
 
--- a/Toolset/palettes/inspector/revinspector.livecodescript
+++ b/Toolset/palettes/inspector/revinspector.livecodescript
@@ -99,7 +99,7 @@ on ideInspectObjects pObjects
    palette stack tTargetInspector
    unlock messages
    set the cSelectedObjects of stack tTargetInspector to tSelectedObjects
-   
+   set the cLocked of stack tTargetInspector to tSelectedObjects is not the selectedObjects
    unlock screen
 end ideInspectObjects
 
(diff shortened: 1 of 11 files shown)


montegoulding:
I have marked as blocker because it makes further fixes to two issues we have merged in notes for

montegoulding:
For reference: http://quality.livecode.com/show_bug.cgi?id=18228

Toolset/palettes/inspector/revinspector.livecodescript
```diff
@@ -99,7 +99,7 @@ on ideInspectObjects pObjects
    palette stack tTargetInspector
    unlock messages
    set the cSelectedObjects of stack tTargetInspector to tSelectedObjects
-   
+   set the cLocked of stack tTargetInspector to tSelectedObjects is not the selectedObjects
```
livecodeali: This might have to be revIDESelectedObjects() instead of the selectedObjects - the latter is empty when there is just a focused stack, so this will cause stack inspectors to lock automatically won't it?

livecodepanos:
@montegoulding Could you please add a bugfix note for bug 18047? It seems that the previous PR for bug 18047 did not include one. No need to add a bugfix note for bug 17693

montegoulding:
@livecodepanos done... I think I had just happened to fix a related bug and not really read the report properly but this patch should resolve the remaining issues there now

livecodeali:
This is excellent by the way :-)
@livecode-vulcan review ok 9470c2b

livecode-vulcan:
💙 review by @livecodeali ok 9470c2bec72baebbd7c21a5ef9b0f54581b6f238

montegoulding:
thanks @livecodeali 😄

jameshale:
sorry but shouldn't line 11633 of _revIDELibrary8.livecodescript_  be

  if word 1 of tOwner is among the words of "**group** bkgnd background" then

and NOT

  if word 1 of tOwner is among the words of "**group,** bkgnd background" then

(The comma after the word group is included in the word chunk so it is actually "group,". Perhaps you began thinking "is among the items" and changed your mind?)

montegoulding:
@jameshale indeed yes. This was fixed in [here](https://github.com/livecode/livecode-ide/pull/1350/files)

jameshale:
@montegoulding sorry about that.
Used to seeing fixes to recent merges titled along the lines of "really fix...[old text of merge])
Always brings a smile to my face.

montegoulding:
No problem @jameshale the more eyes on the code base the better!

PR #1349, opened 2016-08-20
[[ Search ]] Faster object tree iteration

montegoulding:


Toolset/libraries/revidelibrary.livecodescript
```diff
@@ -125,14 +125,13 @@ end revIDEShowMessageBox
...
-function revIDEListSearchObjectTree pRootObject, pObeyDontSearch, pMarked, pIncludeSubstacks, pIgnorePasswordProtected, pScriptsOnly, @rList
+function revIDEListSearchObjectTree pRootObject, pObeyDontSearch, pMarked, pIncludeSubstacks, pIgnorePasswordProtected, pScriptsOnly, @xList
    --delete variable sAddedObjects
    
    # Put the root object into the list first. If this is not possible because to dontSearch or card marking then
    # return an error and put empty into the list. What might also happen is that one of the child objects
    # as its dontSearch set to true. In this case, we'll return a partial list.
-   if word 1 of the name of pRootObject is among the words of "field group card" and pObeyDontSearch then
-      put empty into rList
+   if pObeyDontSearch and word 1 of the name of pRootObject is among the words of "field group card" then
```
livecodeali: I presume this should actually check the `dontSearch` property of `pRootObject` too

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -4382,11 +4386,13 @@ private function findListObjectsOnAllAvailableStacks pCallback, pCallbackTarget,
```
livecodeali: You could probably make this slightly more efficient by populating the `tMainstacks` variable with either `the mainstacks` or  `ideUserMainStacks()` depending on `gRevDevelopment`, then both the conditions in the repeat loop could go.

I wonder if it should be using `ideMainStacks()` instead anyway, which does the above but based on the `cShowRevolutionStacks` preference instead of `gRevDevelopment`. What do you think?

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -4455,9 +4457,12 @@ private function findListObjectsOnAllStacksAndSubstacks pCallback, pCallbackTarg
    
    local tList
    repeat for each line tMainstack in tMainstacks
+      if there is not a stack tMainstack then
```
livecodeali: Is this actually possible?

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -4455,9 +4457,12 @@ private function findListObjectsOnAllStacksAndSubstacks pCallback, pCallbackTarg
```
livecodeali, line 4441: Same comment as above applies for the mainstacks variable. Perhaps we need an `ideMainStacks pIncludeIDEStacks` if returning them based on gRevDevelopment is common.

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
montegoulding: hmm... I think I put it in because I hit an error on a non-existent stack...

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
montegoulding: I think it can be governed by `cShowRevolutionStacks`... not sure why it isn't.

Toolset/libraries/revidelibrary.livecodescript
montegoulding: That's a good question... I only reversed the conditions here because the boolean check would be quicker and didn't apply much other brain power to the condition ;-)

Yes I think it should check that...

livecodeali:
@livecode-vulcan review ok 4235a6e

livecode-vulcan:
💙 review by @livecodeali ok 4235a6e3eb4024c31306e48d9a5d67875e2db06c

montegoulding:
@livecodeali what are your thoughts on a target for this?

PR #1352, opened 2016-08-20
[[ PubSub ]] Implement optional mutable params on messages

montegoulding:


livecodeali:
What was the use-case for this again?

I'm a bit worried about the fact that there is no ordering to the subscriber list, so that you don't know who's had a shot at mutating the parameter before you receive the message.

montegoulding:
Hmm... maybe we need some way to give subscribers the original parameters and the current state but the general idea is you break it and you get to keep both pieces. I'd like to use this for menus and standalone build settings for a start.

montegoulding:
So use case:
1. build menu
2. let subscribers mutate the menu
3. show the menu
4. handle menu items
5. fall through to publish menu item chosen

PR #1353, opened 2016-08-21
[[ BracketCompletion]] Implement bracket completion

montegoulding:


montegoulding:
The first two commits of this PR will be included in a forthcoming script formatting issues PR

livecodeali:
The bracket highlighting doesn't seem to be quite there yet - I can get it retain the background color between brackets by deleting to just after the first bracket and then carrying on typing, for example

screen shot 2016-08-29 at 10 05 31

I think this is to do with the engine rules for block continuation, you'll probably have to do something to explicitly set the background color of the new chars to empty. 

There are a couple of other little problems:
1. The bracket matching code doesn't take line continuations into account
2. When a pair of brackets highlights the block measurement changes slightly and you get a shift of all the text, which looks kind of bad. I guess this is an engine bug, but I don't know if I'd want to include this without having fixed that bug. What do you think?

PS Just did a little test out of curiosity, `the last line of (char 1 to tCharIndex of field “blah”)` is much faster than `line (the lineIndex of char tCharIndex of field “blah”)` so good that you've used the former everywhere :-)

montegoulding:
Hmm... I didn't notice the shift. I'll have a play with it again. I might pull bracket highlighting out of this PR so the completion can be dealt with separately if there's engine bugs that can hold it up.

livecodeali:
I'm also getting debugger library errors on this branch when I create a stack, open the SE and add some text, then delete it

montegoulding:
ok, I'll check that... I'm having a little play with it now and I can't see the block measurement issue. Would it be a font thing?

montegoulding:
I can't replicate your errors but I've made a few alterations

montegoulding:
I've committed my current state to see if I've fixed the errors if you want to pull it

livecodeali:
Might be a font thing, yeah...

montegoulding:
Nope, can't see any text shift during highlight. I tried a few fonts and resized to 18pt.

livecodeali:
@montegoulding Just realised the error is in develop, not this branch - just press backspace in an empty script

montegoulding:
Aha! we'd better check if that's in RC 1

livecodeali:
Doesn't seem to be

livecodeali:
demo

montegoulding:
ew... what font is that?

livecodeali:
default on Yosemite, so Helvetica Neue-ish?

montegoulding:
OK, that looks like it would be really curly to resolve.. 😞

livecodeali:
We could maybe add something primitive in the SE to check if the font being used is fixed width?

montegoulding:
Why isn't on and end bold? I can't get this to happen on El Capitan so perhaps it's a CoreText bug

livecodeali:
Because the default font on Yosemite doesn't have a bold style

montegoulding:
oh... does this only happen if you use that font?

montegoulding:
highlight

montegoulding:
That's with `(System)` on El Capitan as that's the only one that doesn't have bold... they all work fine

livecodeali:
@livecodefraser points out it's because of the kerning rules - any font where the gap between ( and another letter is reduced

montegoulding:
well should we just suggest people use a font that doesn't have that ;-)

peter-b:
@montegoulding I don't think that's a blocker.  It's pretty normal for fonts to change character widths in bold styles.  I think it's probably okay to ship this as it is, but there are two alternative options:
1. Make the script editor default to a suitable platform-dependent fixed width font
2. Remove the bold effect and just use a coloured background to highlight the matching delimiters

montegoulding:
@peter-b actually I'm not bolding the brackets. I'm just setting `the backgroundColor` of them. We could probably include an open source cross platform font designed for code like [this](http://sourcefoundry.org/hack/)

peter-b:
@montegoulding Could we bundle Adobe Source Code Pro?

peter-b:
https://github.com/adobe-fonts/source-code-pro

montegoulding:
sure... just downloading now

montegoulding:
source code pro looks good... would be nice to bundle it and load as we do with fontawesome

peter-b:
@montegoulding Just tested these changes in the current `develop` IDE.  I can get the highlighting of matching brackets to work, but when I select some text and type a <kbd>(</kbd> it just replaces the selection rather than wrapping it.

montegoulding:
PR #1619 replaces this

PR #1356, opened 2016-08-22
Do not show warning about data loss when saving as latest format

livecodepanos:
Create a stack and click cmd+S. Select latest format ("LiveCode stack") and click the Save button. This caused `tVersion` to be empty, and a warning dialog saying:

screen shot 2016-08-22 at 14 30 27

This patch ensures that no warning dialog is shown if the user chooses to save the stack as latest format ("LiveCode stack").


livecodeali:
@livecode-vulcan review ok 513e6e0

livecode-vulcan:
💙 review by @livecodeali ok 513e6e043166725be355a6efcdd989e22aee1ce2

PR #1359, opened 2016-08-22
[[ SEFormatting ]] Fixes to script editor formatting

montegoulding:


Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -951,7 +951,7 @@ command scriptComment
    # OK-2009-03-03 : Bug 7537 - Reformat the handler after commenting, this may be flawed,
    # but the proper solution is rather time-consuming.
```
livecodeali, line 950: I guess you should adjust these comments 😛

montegoulding, line 950: perhaps.... unfortunately I don't know how OK thought it was flawed. There's probably something I'm missing there.

montegoulding:
@livecodeali could you review again as @trevordevore spotted something else on the merged PR

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
livecodeali, line 950: Commenting and uncommenting multiple handlers only results in the first being formatted, which is why it should be "script" and "handler" is flawed. The proper solution ("script") is adjudged to be too time-consuming. So I think the comment and the equivalent one below should just be deleted.

montegoulding, line 950: @livecodeali I think perhaps he may have been referring to reformatting just the commented/uncommented lines which my patch doesn't do so it does still apply

montegoulding, line 950: haha... cross posting with different interpretations ;-)

montegoulding, line 950: Now that I change it to script scope commenting out a handler in the IDE library is sloooow... I think this needs some more logic appied

livecodeali, line 950: I think we have the same interpretation... the original version only reformats the 'current handler' (I think the one containing the beginning of the selection). Your change makes it do the correct thing by reformatting everything... or am I wrong?

livecodeali, line 950: Yes, that's what the comment is saying isn't it? Doing the proper thing (i.e. reformatting the whole script) is too slow.

montegoulding, line 950: well it doesn't seem the correct thing if you are only commenting a few lines from a many thousand line script... ideally it would only format the selected lines

livecodeali, line 950: Maybe there should be a "selection" option for `scriptFormat`

montegoulding, line 950: I think we need a `selection` mode for scriptFormat

montegoulding, line 950: that's gold ;-)

montegoulding: @livecodeali after messing with this for most of the day I came to the conclusion that we would need to do a complete overhaul of formatting in order to support just formatting the selection. Accordingly I have updated the PR to support multiple-handler ranges for handler mode.

livecodeali: @montegoulding After mulling it over I think that makes sense anyway, as there might be cases where commenting or uncommenting over the boundaries of two handlers affects the indenting later on.

livecodeali:
@livecode-vulcan review ok d633482

livecode-vulcan:
💙 review by @livecodeali ok d63348211986f61c6ef4155636cde01aa0dc784f

PR #1373, opened 2016-08-27
[Bug 4084] Multiline comments

montegoulding:
This PR depends on https://github.com/livecode/livecode/pull/4425

It provides support for leading whitespace within multiline comment blocks and some basic editing
features (shift left | right, reindent) for when autoformat is disabled or within a multiline comment where it shouldn't apply.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
@@ -2183,6 +2183,7 @@ private command textReplaceRaw pOffset, @pOldText, @pNewText, pDontUpdateBreakpo
    unlock screen
 end textReplaceRaw
 
+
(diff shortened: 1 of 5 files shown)


peter-b:
Is this also related to [bug 4084](http://quality.livecode.com/show_bug.cgi?id=4084)?

montegoulding:
Yes to fix 4084 we need to detect if a line is a multi-line comment or not and we need the same for colorization I've just added an _internal command to get the comment nesting for the line but the colorization still isn't quite right

peter-b:
Okay thanks. It would be great to get this fixed.

montegoulding:
it actually has the release note already

peter-b:
@montegoulding Yes but I wanted to put it in the title so I can more easily see what I need to poke in BZ when it gets merged. ;-)

montegoulding:
yes sorry I started working on it then I searched BZ after to try and find related reports

montegoulding:
cleared the milestone because multiline comments detection needs more work

PR #1375, opened 2016-08-31
[[ Bug ]] Fix potential error on stack deleted in tight loop

montegoulding:
I have set the milestone to 8.1.1 because this is a very low risk patch

### DIFF
--- a/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
+++ b/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
@@ -68,6 +68,10 @@ on ideNewStack
 end ideNewStack
 
 on ideOpenStack pTarget
+   if not exists(pTarget) then
+      exit ideOpenStack
+   end if
+   
    if the short name of ideStackOfObject(pTarget) is not the short name of me then
       startCenterClose
    end if


livecodeali:
@livecode-vulcan review ok ee5c471

livecode-vulcan:
💙 review by @livecodeali ok ee5c4719e9fbcd2213892639430e613ceabdbdca

peter-b:
- Is it possible to add a test to this patch?
- Does this problem affect any other IDE messages?

montegoulding:
@peter-b it's easy enough to test in the IDE so I suspect I can devise a test. Essentially `msg` should remain empty if `gRevDevelopment` is true when you create and delete a stack in a tight loop with messages unlocked. Probably the test could cover other objects also. It could feasibly impact any object that is sent in time with a target parameter as the time will allow time for the handler that triggered the message to delete the target.

montegoulding:
@peter-b thinking about the test to truly test the IDE I would need to have a list of all the controls that subscribe to a particular message and get them to load in their correct context... It's easy enough to test the out of the can IDE but harder to test every subscriber and get those stacks into the correct state that they would be subscribing in...

montegoulding:
@peter-b I was expecting the following test to fail without this patch so perhaps there's some detail I'm missing that was causing me to hit this error while working today:

```
on TestObjectCreationAndDeletionInOneHandler
   global gRevDevelopment
   put true into gRevDevelopment
   revIDEOpenPalette "Start Center"

   put empty into msg
   create stack "test"
   create button "test"
   delete button "test"
   delete stack "test"
   wait 10 milliseconds with messages
   TestAssert "stack deletion doesn't generate errors", msg is empty
end TestObjectCreationAndDeletionInOneHandler
```

PR #1380, opened 2016-09-07
[Bugfix 15830, 13997] Improve user feedback if breakpoint conditions are invalid

montegoulding:
This patch moves the scripts for the breakpoint pane and the breakpoint editor stack into behaviors. It also removes the breakpoint condition editor and reuses the breakpoint editor from the context menu in the script gutter.

User feedback is improved using:
- a compile check on the condition
- colorisation of the of the condition field while typing

Additionally the condition field now also uses the script editor font and background color settings.


peter-b:
@livecode-vulcan review ok 9f2c1d8

livecode-vulcan:
💙 review by @peter-b ok 9f2c1d89bc6695c4b285ed95381ea1bb44f22f3a

PR #1387, opened 2016-09-16
[[ Bug 6937 ]] Implement code folding in script editor

montegoulding:


Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -5424,3 +5455,358 @@ private command EditorFieldCacheSwap pOldRuggedObjectId, pNewRuggedObjectId
...
+   local tEndBlock
+   local tCounter
+   
+   put word 2 of the clickLine into tLine
+   put line tLine of field "Script" of me into tLineText
+   
+   -- handlers beginning with on, function or command
+   if word 1 of tLineText is among the items of   "on,private,function,command,setProp,getProp,after,before" then
+      if word 1 of tLineText is "private" then
+         put trueWord 3 of tLineText into tHandlername
```
montegoulding, line 5473: this should probably use token rather than trueWord

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -5424,3 +5455,358 @@ private command EditorFieldCacheSwap pOldRuggedObjectId, pNewRuggedObjectId
    -- controls.
    __EditorFieldCacheFlushExcess
 end EditorFieldCacheSwap
+
+on folding
```
montegoulding, line 5459: private command __Uppercase (a number of places)

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -5424,3 +5455,358 @@ private command EditorFieldCacheSwap pOldRuggedObjectId, pNewRuggedObjectId
...
+   local tHandlerLastEnd
+   local tHidden
+   local tBeginU
+   local tEndU
+   local tScriptTemp
+   local tNumbersTemp
+   local tUnderlineText
+   
+   put " hidden" into tHidden
+   put "<u>" into tBeginU
```
montegoulding, line 5738: I wonder if folding to something like this would be better than the underline thing:

```
command MyCommand
end MyCommand
```

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -5424,3 +5455,358 @@ private command EditorFieldCacheSwap pOldRuggedObjectId, pNewRuggedObjectId
...
+      put word 4 of aLine into tHandlerEnd
+      put line tHandlerLastEnd +1 to tHandlerStart - 1 of tScriptHTML & return after tScriptNewHTML
+      put line tHandlerLastEnd + 1 to tHandlerStart of tNumbersHTML & return after tNumbersNewHTML
+      put line tHandlerStart to tHandlerEnd of tScriptHTML into tScriptTemp
+      put line tHandlerStart + 1 to tHandlerEnd of tNumbersHTML into tNumbersTemp
+      put line 1 of tScriptTemp into tUnderLineText
+      
+      put char 1 to 3 of tUnderlineText & tBeginU & char 4 to - 5 of tUnderlineText & tEndU & char -4 to - 1 of tUnderlineText & cr  after tScriptNewHTML 
+      delete line 1 of tScriptTemp
+      repeat for each line aLine in tScriptTemp
```
montegoulding, line 5766: `aLine` here shadows `aLine` in the parent iterator. Could cause bugs down the track

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -5424,3 +5455,358 @@ private command EditorFieldCacheSwap pOldRuggedObjectId, pNewRuggedObjectId
...
+   
+   put "<p>" into tStart
+   put "</p>" into tEnd
+   put "<p hidden>" into tStartHidden
+   
+   put the styledText of field "Script" into tScriptA
+   put the keys of tScriptA into tKeys
+   sort tKeys ascending numeric
+   
+   repeat for each line aKey in tKeys
```
montegoulding, line 5802: Perhaps we should just be mutating the styledText here as I expect it would be significantly quicker

BerndN: I would have loved to do that but
Bug 18403 - [styledText] Hidden property of a line is not transferred with styledText
i.e. if you set the hidden to true in the styledText and reapply it it will not work.
And in my experience doing html on the fly is a tad faster than working on the styled text.

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: It certainly would be faster, however for "case" it would not work since there is no "end case" and not necessarily a "break"

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: aggreed

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: I used trueWord because I found one instance in a getProp handler: getProp xyz[abc] 
I did not try token though.

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
montegoulding: See https://github.com/livecode/livecode/pull/4526

montegoulding:
Close see #1970

PR #1389, opened 2016-09-17
Uninstall can lead to deleted files if uninstall is part of installation operation

trevordevore:
Uninstall was using `send in time` during the process. During installation of an extension, uninstall is called. Since `send in time` is used the installation could finish before uninstall finished. When uninstall did finish it would delete the extension folder that had just been installed.

### DIFF
--- a/Toolset/libraries/revideextensionlibrary.livecodescript
+++ b/Toolset/libraries/revideextensionlibrary.livecodescript
@@ -466,7 +466,7 @@ on __extensionUninstallCheckInUse pCacheIndex
    __extensionSendProgressUpdate pCacheIndex, "Checking if extension is in use", 20
    
    # Next step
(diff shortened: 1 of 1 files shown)


montegoulding:
@trevordevore this seems OK to me but perhaps @livecodeali can shed some light on why these handlers were sent in time before we merge it in. If they are needed then an alternative would be a callback at the end of the uninstall to continue the install.

livecodeali:
I don't know why they were done this way in the first place - I'm pretty sure there was no especially good reason.

montegoulding:
That's good enough for me @livecodeali ;-)

@livecode-vulcan review ok 4b6e543

livecode-vulcan:
💙 review by @montegoulding ok 4b6e5435cf26223ea3624c65063e87f08f9b7908

PR #1391, opened 2016-09-21
[[ Bug 18292 ]] Fix palettes not showing after being hidden

montegoulding:
I have targeted this for 8.1.2 because it's annoying a lot of people... including me ;-)

### DIFF
--- a/Toolset/libraries/revbackscriptlibrary.livecodescript
+++ b/Toolset/libraries/revbackscriptlibrary.livecodescript
@@ -504,8 +504,6 @@ function revSaveCheck pStack, pNoWarning
             break
       end switch
    else if word 1 of pStack is "revScriptEditor" and the mode of stack pStack is not 0 then
-      -- check script is saved
-      send "revCheckScriptSaved" to cd 1 of stack pStack
       -- this will exit to top on cancel
       lock messages
       close stack pStack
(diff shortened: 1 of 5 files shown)


Toolset/libraries/revbackscriptlibrary.livecodescript
```diff
@@ -2309,42 +2194,105 @@ private function revSEStackIsScriptEditor pStackname
    return false
 end revSEStackIsScriptEditor
 
+on revSEHidePalettes
+   cancel sPalettesMsgID
+   send "revSEDelayTogglePalettes false" to me in 200 milliseconds
```
peter-b, line 2149: This is probably a stupid question, but why 200 ms?

Toolset/libraries/revbackscriptlibrary.livecodescript
```diff
@@ -2309,42 +2245,105 @@ private function revSEStackIsScriptEditor pStackname
...
+   set the lockMessages to tOldLock
    unlock screen
-   unlock messages
-end revSEDelayShowPalettes
+end revSEDelayTogglePalettes
+
+
+private command log pMessage
+   put pMessage & return after msg
+end log
```
peter-b: Can this `log` handler be removed?

Toolset/libraries/revbackscriptlibrary.livecodescript
montegoulding: @peter-b No particular reason other than it allows time for other scripts to toggle the palettes shown and hidden multiple times without doing anything and still presents the palettes or hides them depending on the final state within an imperceptible time for the user.

Toolset/libraries/revbackscriptlibrary.livecodescript
montegoulding: @peter-b It looks like it can be factored out so I will do that... I'll add another commit to this patch that does further cleanup

peter-b:
@livecode-vulcan review ok d83ab59

livecode-vulcan:
💙 review by @peter-b ok d83ab59f9cbea43380e6f334ed9b55f29deffbc1

PR #1394, opened 2016-09-21
[Bug 18452] Make sure saving a substack from the PB saves the mainstack

livecodepanos:
The `revIDESaveStack` is called when you click "Save" on a stack/substack from the menubar. Use this function when saving a stack/substack from the PB, since it handles correctly the case of substacks


peter-b:
@livecodepanos It would be nice to add a "Save Substack As" option to the context menu of substacks, maybe.  Should I file an enhancement request?

peter-b:
@livecode-vulcan review ok 69c5e9f

livecode-vulcan:
💙 review by @peter-b ok 69c5e9f20a469516e08cf09c0dfc88b4384c21b2

PR #1405, opened 2016-09-28
[PI Stack] Add Style property

seaniepie:
To make it easier for a developer to change the style of a stack window to Modeless, Modal or Palette.

### DIFF
--- a/Toolset/resources/supporting_files/property_definitions/com.livecode.interface.classic.Stack.txt
+++ b/Toolset/resources/supporting_files/property_definitions/com.livecode.interface.classic.Stack.txt
@@ -7,7 +7,7 @@ name
 title	
 mainStack								execute:get revIDEMainstackPropertyOptions()			
 decorations												
-style							toplevel|Toplevel,Modeless,Modal,Palette
+style							Toplevel,Modeless,Modal,Palette
 windowShape												
 scaleFactor							1					
 shadow		Basic					true					
--- a/Toolset/resources/supporting_files/property_definitions/com.livecode.interface.classic.Stack.txt
+++ b/Toolset/resources/supporting_files/property_definitions/com.livecode.interface.classic.Stack.txt
@@ -7,7 +7,7 @@ name
 title	
 mainStack								execute:get revIDEMainstackPropertyOptions()			
 decorations												
-style												
+style							toplevel|Toplevel,Modeless,Modal,Palette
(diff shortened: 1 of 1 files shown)


livecode-vulcan:
📝 Hi @seaniepie, I haven't been able to verify that you've signed our [LiveCode Contributor's Agreement](https://livecode.com/account/developer/contribute).

If you have previously signed the Contributor's Agreement, I may not be able to detect it because you haven't [linked your GitHub account to your LiveCode account](https://livecode.com/account/developer/github).

Please see the [information for contributors](https://github.com/runrev/livecode/blob/develop/CONTRIBUTING.md) for more information.

livecodeali:
@seaniepie Since the style property has different values for each object type, you will need to add the options for it. So the line needs to be

```
style<tab><tab><tab><tab><tab><tab><tab>toplevel|Toplevel,...
```

Unfortunately we don't yet have a way to instrospect on properties to get their possible values, otherwise we could essentially do this automatically.

seaniepie:
Thanks Ali, I saw that eventually last night. I was trying to see if anything to do with Modes or other window properties might help. It was just a thought as I was trying to create a drawer or palette window without using a script each time to change it. We used to be able to do this easily pre v7. Of course the Drawer command does not work as it used to now that Yo-se-mite and Ford Sierra XRi have got rid of window drawers. 

Ultimately, lets close this idea for thinking about another time.

livecodeali:
@seaniepie You don't have to close the PR - just tweak the line so it lists the options and it will work!

livecodeali:
@seaniepie Pretty much, although the format for the options is actually:

```
<value>|<displayValue>,<value>|<displayValue>,...
```

My example was misleading, because we actually don't need to distinguish value and display value since they are the same in this case so just

```
Toplevel,Modeless,Modal,Palette
```

will suffice. (i.e. you can delete that initial `toplevel|`. The default value is actually a separate column but I don't think it needs to be added here.

seaniepie:
I didn't understand you at first but I get it now.

livecodeali:
@livecode-vulcan review ok 74c102b

livecode-vulcan:
💙 review by @livecodeali ok 74c102b4133cc4f81feb0691aa944bbea707df0a

peter-b:
@seaniepie Thank you for the patch!

PR #1419, opened 2016-10-05
Permit iOS standalone deployment with two-digit minimum version

peter-b:
Corrects a string parsing error where it was assumed that the first
two components of an iOS version would be exactly three characters.


livecodepanos:
@livecode-vulcan review ok 98d971a

livecode-vulcan:
💙 review by @livecodepanos ok 98d971ab05cca417b88c2a736317288c036517c7

PR #1421, opened 2016-10-07
[Bug 18528] Find Next/Previous in the SE respects current cursor location

livecodepanos:
Called `get the selectedChunk` in the correct place so as to properly update the `sLastSelectedChunk`, that keeps track of the starting point from which the Find Next/Previous should begin.

Setting this to blocker for 8.1.1 RC-2 since this bug was introduced in 8.1.1 RC-1


peter-b:
@livecode-vulcan review ok f106b1b

livecode-vulcan:
💙 review by @peter-b ok f106b1b32ce8fed63f0cb38afd83d29fbb5a7e24

PR #1425, opened 2016-10-10
Set Standalone Settings minimum iOS version to 8.3

seaniepie:
Livecode 8.1.x supports only from iOS v8.3 and later.
- Modifications to 'label of button "Minimum iOS Version"' and 'tSettings["ios,minimum version"]'
- Modification to 'updateIosSettingState' 'set the enabled of button "32bit_only"' to be false as this will no longer be needed
- Minimum iOS variant button on IDE set to have 8.3 as the minimum and 10.0 as the highest


seaniepie:
Sorry if I've overstepped my mark here. I wanted to give it a go more than anything else as I haven't tried playing with the IDE and LC scripts on GIT so far. Let me know if it's any good and where I need to do things better.

montegoulding:
@seaniepie It's good to have a play. Git branches and GitHub PRs are cheap and disposable ;-)

We actually support a minimum deployment target of iOS 6. Recently bumped up from 5.1.1 for Xcode 8 as it broke 5.1.1 support. There's a bit of confusion around deployment targets and SDKs things are built against. The deployment target is the minimum version you want the app to run on while the SDK (what's in Xcode that we compile against) builds slightly different binaries for each version we build against and as externals must be statically linked we need to do that with binaries built with the same SDK.

seaniepie:
It's interesting coz I just had to rework a job for a client on an old iPad2 running iOS 7.1.2. So I pulled in the old stack from a year ago which I had already 'modernised' to the new sockets you'd put together based on your old mergSockets. The profiles had ran out so the app wouldn't open and it wasn't as easy as just load in some new profiles. 
Here's the weird thing (on LC8.1.1rc1). If I set my app up for 7.1 right away then build it tells me at the end of the whole process (as usual, I wish it was at the start) that it can't build it for 7.1 with this version and it should be for 10.0. But if I set it for 9.3, build for that and _then_ set it to 7.1 it works no problem. I've got it working now and that's all cool for my client but I'm going to have to record my findings and submit a bug report for this I think coz it's just too weird.
Thanks for your advice Monte. It's good to get the input. 😄

montegoulding:
That does sound like a bug so it would be good if you can come up with a good recipe for it.

livecodeali:
@seaniepie Definitely not a case of overstepping your mark! We do have an issue with changes to binary stacks though - we can't really accept outside changes to them at the moment because we have no way of checking exactly what has been changed like we do with script-only stacks or other text-based source files. 

It's also tricky from the point of view of branching. The problem is particularly evident with the standalone settings stack, which often requires release-specific changes and causes all sorts of issues when trying to merge various branches!

seaniepie:
Thanks @livecodeali . If I were to see something on a stack then would it be best to provide with it a difference chart to display what has actually been changed?
In this instance I won't be progressing it any further because as we've discussed, although support is not provided for iOS <8.3 LC still seems to build semi-happily for even 7.1 and maybe further.

livecodeali:
@seaniepie The only way to do it would be to tell us exactly what changes were required and get us to implement them, unfortunately, as there's no way to guarantee the changes you've made are the ones you've provided.

PR #1428, opened 2016-10-11
[Bug 18485] Prevent relayering grouped controls unless in Edit Group mode

livecodepanos:
This is achieved by setting the `relayerGroupedControls` to false in the appropriate place, which matches the LC 7.x behavior.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -9062,7 +9062,7 @@ on revIDERelayerControls pObjectList, pMode
    local tRelayerGrouped
    put the relayerGroupedControls into tRelayerGrouped
    
-   set the relayerGroupedControls to true
+   set the relayerGroupedControls to false
    switch pMode
       case "send to back"
          sort lines of tLayers numeric descending
--- a/notes/bugfix-18485.md
+++ b/notes/bugfix-18485.md
@@ -1,0 +1,1 @@
+# Prevent relayering of grouped controls if we are not on Edit Group mode


peter-b:
@livecode-vulcan review ok 39b98a2

livecode-vulcan:
💙 review by @peter-b ok 39b98a2846b77318506ba3734e4ca510d63c3324

montegoulding:
This will result in uncaught execution errors I think... that and some layer changes succeeding leaving things in a partially completed state.

peter-b:
@montegoulding Do you have any idea what the Right Thing is here?

montegoulding:
You have two choices I think:
- easy - iterate the controls to ensure they all have the same owner and error out before touching anything if they don't
- harder - relayer controls within their owners (probably using the relayer command is easier to do that) and relayer owners if required. This is probably what the user wants if they are relayering for something like field tab order which is based on layer.

livecodepanos:
What if I put  the command `set the layer of tObjectArray[tLayer] to ..` in a `try-catch` block? This is what LC 7 does.

peter-b:
@livecodepanos I think @montegoulding was saying that if you do that it might result in some controls being relayered and others being left unmodified, without any clear indication to the user or explanation about what happened.

montegoulding:
@livecodepanos @peter-b thinking on this some more I think the best approach is to:
1. ensure all things being relayered have the same owner which quite easy by iterating the controls and checking the long id of the owner is the same
2. ensure relayering occurs within the owner which is also relatively easy but you do need to be careful:
   - set the relayerGroupedControls to true
   - ensure you never set the layer of one of the controls to <= the layer of the owner or > the layer of the owner + the number of controls of the owner (this will ensure you don't relayer a control out of a group).

livecodepanos:
I had another look at this and I think the bug here is that in LC 8 you can change the layer of a grouped control from the menubar (Object -> ..) **without taking into account the value of relayerGroupedControls**.

So I think it is incorrect to `set the relayerGroupedControls to true` in this handler, since this prop should only be set by the user. 

If you agree, I will remove the `set the relayerGroupedControls to true` line from this handler, and place the lines `set the layer of tObjectArray[tLayer] to ..` in a `try-catch` block.

In this case, if the user has set `relayerGroupedControls` to true then the relayering will occur as expected. We don't need to check that the new layer will be within the range of the group's layers, since according to the dictionary:

_if the relayerGroupedControls property is set to true, and you set the layer of a grouped control to a number greater than the topmost control in the group, or lower than the bottom-most control in the group, the control is moved out of the group_ . 

If `relayerGroupedControls` is false and the user tries to relayer grouped controls, an error can be presented to the user (in the `catch` block).

WDYT?

montegoulding:
Hmm... I think it is probably unexpected behavior for the IDE to move objects out of a group regardless of the value of `relayerGroupedControls`. It is certainly helpful for the IDE to be able to relayer objects within a group without having to go into the awkward group edit mode.

Looking at the original report I think all that needs to happen is to guard against moving a grouped control out of its group. We can do this by either not relayering it outside the owner bounds or relayering the owner when the grouped object needs to be relayered beyond the bounds of the group.

livecodepanos:
Closing this as no work has been done for a long time

montegoulding:
@livecodepanos do you want me to have a look into this?

livecodepanos:
@montegoulding sure feel free to take this

PR #1440, opened 2016-10-13
[Bug 18546] fix the GrayNN values in the text files

mwieder:
This pull request fixes colorsToNames.txt and namesToColors.txt to bring them up to parity with rgb.cpp. I have unit tests for these, but I don't know where to put them. On my local system I have them in the livecode repository with the graphics tests because I can run them there. Any ideas welcome.


peter-b:
@mwieder Thanks for taking a poke at this! I wonder if it would be feasible to get rid of either `colorsToNames.txt` or `namesToColors.txt` entirely, because one of them can be generated directly from the other.

Actually, it just occurred to me that the mapping could be generated by parsing the C++ source file.  That would fix the problem _for all time_ by making it impossible for them to get out of sync again. Is that something you're interested in investigating?

livecodeali:
@mwieder I would add to the IDE tests and use `revIDENamedColorToRGB` and `revIDERGBToNamedColor`. You could use Bernd trick from here http://quality.livecode.com/show_bug.cgi?id=16036 to get the engine's actual mapping for each of the colorNames, check `revIDENamedColorToRGB` returns the same value, and then check `revIDERGBToNamedColor` returns the original color name (or maybe that's exactly what you are already doing)

livecodeali:
@peter-b http://forums.livecode.com/viewtopic.php?f=66&t=28044 😄

peter-b:
@livecodeali Oh, haha. Sounds like we're all on the same wavelength, then! Let's stick with the legacy files for now.

mwieder:
@peter-b in the long run, sure. In fact that's how I originally wanted to approach this. I'm just not sure what would be required to have gyp do the building.

@livecodeali the tests I wrote actually do a round(grayValue/255*100) as one step of the verification, and ensuring that item 1 == item 2 == item 3 as a secondary test.

How do you get ide tests to run? make check in the ide repository doesn't do the trick.

peter-b:
@mwieder You'll want to `make ide-check` in the `ide/tests` directory.

mwieder:
peter- I tried that several times today. It runs partway, then just hangs. Even without my tests. I left it for several hours just to see if it would return.

montegoulding:
Thats the same as an old lib I have:

```
set the visible of stack "libColor" to true
  repeat for each line tColor in the colorNames
    set the backColor of cd 1 of stack "libColor" to tColor
    set the backPixel of cd 1 of stack "libColor" to the effective backPixel of cd 1 of stack "libColor"
    put the backColor of cd 1 of stack "libColor" into tColorNameReference[tColor]
  end repeat
  set the customProperties["cColorNameReference"] of stack "libColor" to tColorNameReference
  set the visible of stack "libColor" to false
```

peter-b:
@mwieder Yes, there's an issue there that's proving _very_ resistant to being fixed.  We've been struggling with it for weeks now. 😨

livecodeali:
@mwieder It should work if you do 

```
linux-x86_64-bin/livecode-community -ui ide/tests/_testrunner.livecodescript
```

or whatever the path to the development engine is.

BerndN:
function getRGB pColorName
   local tRGB
   set the backColor of the templateGraphic to pColorName
   set the backPixel of the templateGraphic to (the effective backpixel of the templateGraphic)
   put the backColor of the templateGraphic into tRGB
   return tRGB
end getRGB

livecodeali:
`reset the templateGraphic` 😛

BerndN:
no worries, breaks anyways

http://quality.livecode.com/show_bug.cgi?id=18578

:)

livecodeali:
Ah! It's because they were put in numeric order, which is not the correct order for the search algorithm.

mwieder:
@livecodeali thanks for fixing the order in rgb.cpp - that's the next thing I was going to look into. Nice to know it's a simple fix.

livecodeali:
@mwieder No worries 😄  I had the ulterior motive of wanting to write a C++ test having not written one yet!

mwieder:
Tests now in place.
@livecodeali - revIDENamedColorToRGB() returns valid values, but revIDERGBToNamedColor() doesn't return anything to me. Is there some trick to using it? Where do these functions live?

livecodeali:
Um... LOL?
https://github.com/livecode/livecode-ide/pull/1443

@mwieder That should answer both those questions 😄

mwieder:
@livecodeali Ha! OK - that explains a lot. But in that case using that fumction to verify the contents of the text files sounds a bit self-referential. And if that's the only place the text files are referenced then maybe using functions in rgb.cpp as the source of color information would be better.

livecodeali:
@mwieder That's what I mean - the code @BerndN posted is a way of getting the engine value of each color name, so the test I've suggested would genuinely test that
- the mapping from names to colors in the text file is correct
- the mapping from colors to names in the other text file is consistent with the first mapping

I think that covers all of the functionality that's expected doesn't it? I think the test that eg Gray10 is actually 26,26,26 can go in the engine repo, again using the graphic object trick

mwieder:
@livecodeali well, there are a few things here.

Checking the correct value in rgb.cpp would, as you noted, be a test in the engine repo and therefore be out of the scope of this pull request. I have now implemented a variation on Bernd's test to ensure that the values in the text files match those coming from the engine.

Secondly, using revIDENamedColorToRGB() and revIDERGBToNamedColor() to validate the contents of the text files isn't very rewarding, since both functions use the text files as the authoritative source. It would be just a circular test.

And this pull request is really about fixing the GrayNN values, not about validating the entirety of both text files.

More importantly, though, what I was trying to get at before is the question of whether the text files are necessary at all. Are those two functions actually used anywhere in the IDE or are they just in place as handy tools? If that's the case, I'd like to suggest _not_ getting the values from the text files but doing something similar to Bernd's function instead.

I've seen the suggestion (somewhere on the forums, I think) that the text files are necessary for LCS scripts, although I don't understand why. I'd like to eliminate the need to rely on the text files and get these values from the engine for speed and maintainability.

That said, I'm pushing the latest changes (two problems with the text files uncovered by the tests, and the expanded tests themselves)

livecodeali:
@mwieder On the contrary - the text files are an implementation detail so in fact there is _only_ point in testing the behavior of those functions and not the text files directly. That way, two things are ensured: 
1) The text files always match the engine values
2) Removing the text files and replacing them with a better implementation doesn't invalidate the tests, indeed the tests ensure the new implementation is correct.

As I said in the forum post linked above: "The mapping from engine color names to rgb values and back again in the text files is only used for translating to and from RGB values when interfacing between LCS and LCB". The reason that is necessary is because of this: http://quality.livecode.com/show_bug.cgi?id=16036.

Anyway, yes, I believe the implementation of `revIDENamedColorToRGB` could be replaced by Bernd's code and the `namesToColors` text file removed. I'm not sure the same is true of `revIDERGBToNamedColor` unfortunately 😞

mwieder:
@livecodeali- Thanks. I'll think about reworking revIDENamedColorToRGB for a future PR. I really dislike the idea of having to use the text files as a shim for LCS.

But I'm still missing the point in your first argument, or maybe we're talking past each other. Since the implementation of the revIDENamedColorToRGB function calls __fetchNamesToColors and the revIDERGBToNamedColor function calls __fetchColorsToNames, and each of those handlers takes the contents of the text files to generate an array; I fail to see how validating the contents of the text files against the contents of the text files accomplishes anything useful. And is quite time-intensive, since each call to retrieve a color name or RGB value loads the text file from disk into an array and retrieves a single element from the array. I'll add the tests if you really want them, but I don't see the point.

livecodeali:
@mwieder I do think we must have some crossed wires here somehow.

Testing the text files and testing the IDE API are functionally exactly the same, as the IDE API uses the text files. But the text files are a (private) facet of the implementation of the IDE API, and therefore are the wrong thing to test. The IDE function could be implemented using a custom property, or a private function, or any number of other ways. 

Consider your original test for example. The following two tests currently would do exactly the same thing:

```
on TestGrayNamesToColors
   local tColorNames
   put url("file:" & "../../../Toolset/resources/supporting_files/colors/namesToColors.txt") into tColorNames
   TestAssert "failed to open namesToColors.txt", tColorNames is not empty
   filter tColorNames with "Gray*"
   TestAssert "Gray colors are among color names", tColorNames is not empty
   repeat for each line tColor in tColorNames
      put item 1 of tColor into tName
      replace "Gray" with empty in tName
      if tName is not empty then
        put round(tName*255/100) into tRGB
            TestAssert item 1 of tColor && "is a valid Gray shade", item 2 of tColor is item 3 of tColor and item 2 of tColor is item 4 of tColor
        TestAssert item 1 of tColor && "is the correct color", item 2 of tColor is tRGB
      end if
   end repeat
end TestGrayNamesToColors
```

```
on TestGrayNamesToColors
   local tColorNames
   put the colorNames into tColorNames
   filter tColorNames with "Gray*"
   TestAssert "Gray colors are among color names", tColorNames is not empty
   repeat for each line tColor in tColorNames
      put tColor into tName
      replace "Gray" with empty in tName
      put revIDEColorNameToRGB(tColor) into tRGB
      put round(tName*255/100) into tValue
      TestAssert tColor && "is a valid Gray shade", item 1 of tRGB is item 2 of tRGB and item 2 of tRGB is item 3 of tRGB
      TestAssert tColor && "is the correct color", item 1 of tRGB is tValue
   end repeat
end TestGrayNamesToColors
```

The first one explicitly accesses the text file, the second one uses the IDE API.

Obviously if the text files are changed, both the tests still work. But if the implementation of `revIDEColorNameToRGB` changes, only the second test will continue to test what actually needs to be tested, namely the result of the actual public function that is used.

Also, if the location of the text files had to be changed, then only one place needs to be updated (the IDE) rather than both the IDE and the test.

The IDE function isn't inefficient, it gets the text file once and stores in a local variable.

I really don't mind if you only want to add tests for the Gray numbers, but the test needs to use the IDE function rather than fetching the text file directly.

mwieder:
@livecodeali Ah! That makes sense. I thought you were suggesting something else. Yes, I'll make that conversion, and since I already have tests for all the colors in the text files, I'll keep that in place here.

BerndN:
Jan Schenkel in 2013 did a comprehensive speed test for ColorNamesToRGB on the use-list

[http://runtime-revolution.278305.n4.nabble.com/How-do-I-find-the-rgb-values-for-colors-defined-by-their-name-e-g-blue-green-yellow-etc-td4673481.html#a4673566](url)

It turns out that 

 set the colorOverlay["color"] of the templateGraphic to tColorName 
 put the colorOverlay["color"] of the templateGraphic into tColorRGB 

seems to be the fastest way, also it does not need the effective backgroundpixel.

livecodeali:
@BerndN Nice!

@mwieder It's just occurred to me that we can get rid of the other text file just by populating the colorsToNames array on startup, 

```
repeat for each line tColor in the colorNames
  put tColor into sColorsToNamesA[revIDEColorNameToRGB(tColor)]
end repeat
```

mwieder:
OK - I refactored the tests and they're passing locally. Once I expanded the tests to cover all colors there were some interesting findings: Gray100 is the same as White, Gray0 is the same as Black, etc. so I added a function to disambiguate the aliased colors.

For some reason when I tried to use the templateGraphic I segfaulted every time I tried to set a property. I could read them with no problem, and resetting the templateGraphic didn't seem to have problems, but even if I set the same value into either the colorOverlay["color"] or the backColor or backPixel I ended up with a segfault. I'm not sure yet whether this is the result of running in a unit test environment, because trying exactly the same code in a button script had no problems.

mwieder:
I have another branch building on this that eliminates the need for the text files, based on the discussion above. As far as I can tell it's working properly, although the tests fail with an "lcs check" message no matter whether I test my branch or the develop branch, so it's a bit hard to be sure. Should I just close this PR and submit the new one?

livecodeali:
@mwieder Sorry for the lack of response, I've been desperately trying to get test builds working again after several things happened to them at once!

Yeah I think probably just close this one and submit a new one is the best plan. Hopefully by the end of today the tests should pass again.

PR #1453, opened 2016-10-20
[Bug 18637] Fix searching in "Stack file and its stackfiles"

livecodepanos:
Based on the changes in PR https://github.com/livecode/livecode-ide/pull/1349,
`tList` is no longer a `CR` separated list (it is now an array), so we should not `delete the last char of tList`. 

In the future we might want to rename all array vars to end with `A`, i.e. `tMyCollectionOfStuffA`


livecodeali:
@livecode-vulcan review ok 89ebe68

livecode-vulcan:
💙 review by @livecodeali ok 89ebe68c8f77b63c3448d1e058f356cec69daab4

PR #1454, opened 2016-10-20
[[ Bugfix 18300 ]] Property inspector custom property list is not sorted

trevordevore:
The custom property UI in the property inspector was always sorting keys numerically.
This update changes the sort order based on the keys being displayed. If they are
numeric then the sorttype is set to numeric. Otherwise it is set to text.


livecodeali:
@trevordevore Would you mind retargeting this at develop-8.1 please?

livecodeali:
Oh, you might have to actually rebase it...  :frowning_face: 

@livecodepanos points out that there are two consecutive `of`s in the code

trevordevore:
@livecodeali I was going to rebase it but then I just decided to create a new PR. It was faster :-)

PR #1455, opened 2016-10-20
[[ Bugfix 18300 ]] Property inspector custom property list is not sorted

trevordevore:
The custom property UI in the property inspector was always sorting keys numerically.
This update changes the sort order based on the keys being displayed. If they are
numeric then the sorttype is set to numeric. Otherwise it is set to text.

### DIFF
--- a/Toolset/palettes/inspector/editors/com.livecode.pi.customprops.behavior.livecodescript
+++ b/Toolset/palettes/inspector/editors/com.livecode.pi.customprops.behavior.livecodescript
@@ -37,6 +37,11 @@ on editorUpdate
       set the label of button "customPropertySet" of group "Set buttons" of me to sPropSet
    end if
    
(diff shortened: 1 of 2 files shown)


trevordevore:
@livecodeali Here you go. Based on develop-8.1

livecodeali:
@trevordevore Thanks! I mentioned in the other PR but after you deleted it - you have double `of`s in your

```
...sorttype of of widget...
```

I think the engine allows this but probably best to remove if possible!

trevordevore:
@livecodeali Fixed. Yes, the engine does allow it as I tested the code.

livecodeali:
@livecode-vulcan review ok 903f4b6

livecode-vulcan:
💙 review by @livecodeali ok 903f4b62b9be46ebab9e985411d2dc644a47974a

PR #1460, opened 2016-10-28
Feature eliminate color text files

mwieder:
Allow colorname-rgb and rgb-to-colorname translation without the need for the color text files, using rgb.cpp as the single authoritative source of truth. This replaces the attempt to fix the text files in order to close bug report 18546.

Note that there are test files here, but the ide-check tests are failing on my system in some other test (not my own) so I can't currently verify that they will pass on the CI server.


livecodeali:
@mwieder Thanks very much for this! You don't need separate implementations of color name <-> RGB value in the test file - just use the IDE implementations directly. All the IDE initialisation is done before running the IDE tests.

mwieder:
Admittedly some of the tests are left over from the .txt files, but they at least prevent regressions. And the RGB -> colorname tests pointed out the need for disambiguating the colors. You might hope that revIDERGBToColorName(revIDEColorNameToRGB(tColorName)) would return tColorName, but that isn't always the case since there are some examples where there are multiple names for the same RGB value, thus the need to disambiguate the RGB values to a singleton.

livecodeali:
@mwieder Yes, it makes sense to use your `DisambiguateColorName` handler.

But your revIDEColorNameToRGB wrapper is not needed and your revIDERGBToColorName handler reimplements the IDE handler which is not only pointless but means the test is not actually checking the IDE implementation, which is the whole point of the test.

mwieder:
OIC. Yes, you're absolutely right. Forgot about that. I'll fix it up later today.

livecodeali:
@mwieder Thanks so much! It'll be great to get this in to 9.0 dp 2.

mwieder:
pushed.

livecodeali:
@mwieder Superb, thanks!

livecodeali:
@livecode-vulcan review ok af414b0

livecode-vulcan:
💙 review by @livecodeali ok af414b0f3f12d218bc85294ee183cbcc76579a22

PR #1462, opened 2016-11-07
[Bug 18791] Fix PI "list editor" not updating when it should

peter-b:
The PI editor behaviour had some not-quite-correct code for ensuring
that PI rows didn't get rebuilt/refreshed correctly.  When deciding
whether the Nth value should cause an update, the Nth value was being
compared against the (N-2)th value rather than the (N-1)th value.  As
a result, some property modification patterns were causing the PI to
show stale property values.


livecodeali:
@livecode-vulcan review ok 6cd6bc4

livecode-vulcan:
💙 review by @livecodeali ok 6cd6bc492fba8b6e16ee1b01d7c707d54398a19b

PR #1480, opened 2016-11-16
Update com.livecode.interface.classic.Stack.txt

asayd:
Two property names were incorrect.

PR #1481, opened 2016-11-16
Update propertyInfo.txt

asayd:
Two property names were incorrectly spelled.
Made labels more precise.


asayd:
This change has been moved to PR #1480.

PR #1485, opened 2016-11-18
Create behaviors from script only stacks in behavior area of PI

trevordevore:


trevordevore:
I'm not sure about the wording in the menus. Open to other suggestions.

livecodeali:
For the sake of discussion the new menu items are
```
Create behavior from new script only stack
Create behavior using control script and script only stack
```

I have no problem with the first one, the second one may be slightly unclear. I'm not completely sure how to make it better though.

Maybe it should be another submenu:
```
Create behavior from script only stack
   - create new  (or create empty?)
   - use existing control script
```

trevordevore:
@livecodeali What about this?
d0003977-efc3-4565-bda9-c3876f182e28

livecodeali:
@trevordevore That works for me!

livecodeali:
@livecode-vulcan review ok a1d11d2

livecode-vulcan:
💙 review by @livecodeali ok a1d11d22d6b74d9659542b274bc6a529c75ede8c

PR #1489, opened 2016-11-25
[Bug 18037] Use property defaults from metadata for testing widgets

peter-b:
Up until now, the 'Test' button in the 'Extension Builder' stack used
a raw `create widget` command to instantiate the test widget.  This
produced a "raw" widget, with its properties exactly as initialised by
its `OnCreate()` handler.

By contrast, when created from the tools palette, the default values
specified in the metadata for each property are applied.  These values
are intended to ensure a pretty and/or accessible state for the widget
when constructing a user interface interactively.

Although this behaviour wasn't necessarily wrong, because it helped to
verify that `OnCreate()` initialised the widget to a sensible empty
state, it required some contortions to check what the widget would
look like when dragged out of the tools palette.

This patch modifies the 'Test' button so that it creates a test widget
exactly as if it had been dragged out from the tools palette.


livecodeali:
@livecode-vulcan review ok 7eb66cc

livecode-vulcan:
💙 review by @livecodeali ok 7eb66ccd8e3700de8cafa811db6e134aec840d72

PR #1490, opened 2016-11-25
[Bug 18586] Allow any delimiter in card/group/control names

livecodepanos:
Previously, the PB before expanding the contents (i.e. the "children") of an object X, created a list. This list consists of several lines (one per child), and each line has the form:

`control_layer & TAB & control_name & TAB & control_ID`

This caused problems if a `control_name` contained TAB, because e.g. `item 3 of line X` did not guarantee to return the `control_ID`  

This patch addresses this problem by using a multidimensional array instead of multiple lines of tab-delimited items.


peter-b:
@livecode-vulcan review ok 24dd06a

livecode-vulcan:
💙 review by @peter-b ok 24dd06a2c784feefff3464ab6e6e951dca0f2c3e

PR #1502, opened 2016-12-08
Relative file support for stackfiles

trevordevore:
Currently adding a stack to the stackfiles will only convert it to a relative path if the file is located somewhere within the same folder. This commit will always make paths to the stack relative.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -6122,31 +6122,94 @@ on revIDESetGradientProperty pObject, pProperty, pValue
 end revIDESetGradientProperty
 
 private function __resolveStackFilePath pBasePath, pOtherPath
-   -- First absolutize other path (if possible)
-   local tOldFolder
-   put the folder into tOldFolder
-   set the itemDelimiter to slash
+   set the itemdelimiter to "/"
+   put __resolveRelativeFilenameReference(pOtherPath, item 1 to -2 of pBasePath) into pOtherPath
    
-   local tFolderResult
-   set the folder to item 1 to -2 of pOtherPath
-   put the result into tFolderResult
-   if tFolderResult is not empty then
-      set the folder to tOldFolder
-      return pOtherPath
-   end if
-   put the folder & slash & item -1 of pOtherPath into pOtherPath
-   set the folder to tOldFolder
-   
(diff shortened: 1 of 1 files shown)


trevordevore:
To whom it may concern,

I've been working with the stackfiles quite a bit lately. I often add stack files to the `stackfiles` property that are in different folders than the target stack. This means I have to manually set the `stack files` property in the message box so that my stacks are portable.

This update will always assign a relative path. It means that someone wanting to use a full path would have to use the message box. I'm guessing that is less common than wanting to use a relative path. 

Thoughts?

trevordevore:
stackfiles

livecodeali:
@trevordevore This makes sense to me in principle - my only concern is whether this changes the location of a stackfile in a standalone if it was previously specified with an absolute path. I would hope that it doesn't make any difference - I just want to make sure.

livecodeali:
You know what, we should really have some tests for that distinction in standalones: absolute path -> discard folder structure, relative path -> keep folder structure. I will write one!

livecodeali:
Ok they are supposed to end up in the right place regardless (although they actually weren't https://github.com/livecode/livecode/pull/4988) 

I'm worried about the proliferation of file path manipulation implementations. Could you move those handlers into the IDE library? There is a `revIDERelativeStackPath pStack, pPath` which should really be refactored to use your `__makePathRelative` handler.

livecodeali:
@livecode-vulcan review ok 348903e

livecode-vulcan:
💙 review by @livecodeali ok 348903e2831c5e8a0afa9e431d5e0d21208a287a

Brahmanathaswami:
Thoughts?  This comment may be a bit out of scope, but FWIW:

We've been struggling with this also. our challenge has been colloborating. if Dev A has theloader in his branch and addes a file to the respository, he has to update the stack files for his stack... There was some bug where if you changed the default folde somewhere in your code, then when you added a stack file it would use the long path form the ~/drive.  Then if Dev B is working and changes the loaders stack... it now has a different stack file prop than the one "in the other branch"  This was painful enough that Andre added a handler to the init function that fetches a list of files, (well actually paths relative to the main-loader stack) which updates the stack files property when the stack is run... now, if we add any files to the repository and push, we also push an update to the JSON init file which has the new json object/stack files list.

so yes, this area needs work and we need to think collaborative environment.

PR #1530, opened 2017-01-31
[[ Bug 19188 ]] Ensure correct debug context is used when executing in message box

livecodeali:
- If the user has selected a debug context from the dropdown menu in the script editor, this should be the context of message box execution 
- otherwise, use the debug contexts with the chain of contexts leading up to message box execution removed. This chain can vary in length depending on how it was triggered, so replace the magic number of lines with a calculation using 'returnInField' and 'the target'. This means the filtering will work whatever extra handlers are introduced between the returnInField to one of the message box fields.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -11316,6 +11316,10 @@ on ideCompileCheck pScript
    return empty
 end ideCompileCheck
 
+function ideIsDebugging
+   return revTopMostScriptEditor("debug") is not empty
+end ideIsDebugging
+
 private command __removeMessageBoxFromDebugContexts @xContexts
    local tTarget
    put the long id of the target into tTarget
@@ -11350,7 +11354,7 @@ command ideDoMessage pScript, pDebugMode
    try
       if pDebugMode then
          local tContext, tScriptEditor
-         put revTopMostScriptEditor() into tScriptEditor
+         put revTopMostScriptEditor("debug") into tScriptEditor
          -- Ensure a user-selected debug context in the script editor is the first
          -- option for execution context in the message box
          if tScriptEditor is not empty then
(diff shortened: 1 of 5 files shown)


livecodeali:
I tried to add tests for this, and *almost* got it working, but it's a bit too hard given all the sending in time that the script editor does.

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -11327,22 +11349,28 @@ command ideDoMessage pScript, pDebugMode
...
-         # Line 3 is returnInField (message box field)
-         # Line 4 will hopefully be the user code (assuming of course the message box isnt' changed).
-         # Note that this code is called from both the single and multi-line fields, but luckily they both 
-         # have the same number of calls.
-         local tContext, tContextNumber
-         get revDebuggerContexts()
-         put line 4 of it into tContext
-         
+         local tContext, tScriptEditor
+         put revTopMostScriptEditor() into tScriptEditor
```
montegoulding: @livecodeali I think this should this iterate open script editors for the topmost in debug mode. How about adding a `pMode` param to this function?

Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
```diff
@@ -587,7 +587,7 @@ command ideMessageBoxExecuteScript pScript
       put empty into tIntelligenceObject
    end try
    local tDebugMode
-   put revIDEGetPreference("IDEDebugMode") into tDebugMode
+   put the traceStack is not empty into tDebugMode
```
montegoulding: @livecodeali this could be true when profiling... not sure what to suggest here other than again determining if there's an open script editor in debug mode.

Toolset/libraries/revidelibrary.8.livecodescript
livecodeali: Great idea!

montegoulding:
@livecode-vulcan review ok 42226a0

livecode-vulcan:
💙 review by @montegoulding ok 42226a083bbe5baadbc29ea4d7d9bbaf587fcb1d

PR #1538, opened 2017-02-07
[Bug 19152] Show warning if stack new name begins with "rev"

livecodepanos:


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -5236,6 +5236,14 @@ on revIDEPropertySet pObjectList, pProperty, pValue, pLockUpdates
          
          # If stack name is changed, use new name when setting edited
          if tStack is tObject and pProperty is "name" then
+            if char 1 to 3 of pValue is "rev" then
+               answer warning "Using rev in the first three characters of a stack name is reserved for use by the LiveCode development environment and advanced users creating Plug-ins. If you use these characters, your stack may not behave as expected."
```
montegoulding: @livecodepanos I think both of these answer dialogs need to have their messages wrapped in revIDELocalize. It's not really doing much now but I guess it will one day ;-)

Also perhaps have an opportunity to cancel here?

Eventually it would be nice to do away with the rev prefix thing and use `_ideoverride` everywhere but that's a long way off I think.

livecodepanos:
Thanks @montegoulding, I have pushed an update :)

montegoulding:
@livecode-vulcan review ok a26d078

livecode-vulcan:
💙 review by @montegoulding ok a26d078c009ec6c3e898e1748943346fa6b021f9

PR #1544, opened 2017-02-10
[Bug 18549] Make sure 'lock cursor' works in the IDE

livecodepanos:
The `ideMouseMove` is used for setting the cursor to the appropriate 
 resize cursor when the mouse is over a selection handle, so it makes sense only in *edit mode*. It includes an `unlock cursor` command, which prevents any `lock cursor` command to work in a stack script.

This patch fixes this problem by exiting `ideMouseMove` if the tool is not `pointer tool`.


montegoulding:
@livecode-vulcan review ok 3cc55e8

livecode-vulcan:
💙 review by @montegoulding ok 3cc55e8e21651ba2a1ec624ee1e984799f5db69b

PR #1556, opened 2017-02-27
[Bug 19332] Remove setting realPoints custom prop from revRotatePoly

livecodepanos:
The `realPoints` custom prop was set when calling `revRotatePoly` on a graphic. This resulted in the following unexpected behavior:

Use-case:
1. Reset the `points` of a graphic
2. Rotate the graphic by X degrees
3. Reset the `points`
4. Rotate again by X degrees

OBSERVED RESULT: The graphic was rotated by 2*X degrees, since the `realPoints` prop was not reset, so the new rotation started from the existing `realPoints`. This patch removes this custom prop, since:
- it is not used in other places than `revRotatePoly`
- it is not documented
- the user is not expected to reset the `realPoints` too when resetting the `points`

### DIFF
--- a/Toolset/libraries/revcommonlibrary.livecodescript
+++ b/Toolset/libraries/revcommonlibrary.livecodescript
@@ -559,11 +559,7 @@ command revRotatePoly pGraphic, pAngle
     
     lock screen
     lock messages
-    if the realPoints of pGraphic is empty then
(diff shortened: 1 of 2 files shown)


livecodepanos:
This patch is incorrect, the `realPoints` are indeed needed.

(http://quality.livecode.com/show_bug.cgi?id=19332)

PR #1558, opened 2017-02-28
[Bug 19339] Fix "Pending Messages" tab of msg box

livecodepanos:
Some artefacts are shown in the middle of the output field in tab "Pending Messages". This happens because LC tries to count the number of lines of the script of stack "remotedebugger", so as to display it in the blue box in the right. But this stack is password protected, so access to the script is not allowed, so an error occurs while the card is rendered, so the drawing stops and we only see some artefacts.

This patch fixes the problem, by checking if the target of the pending messages is a password protected stack, and if this is the case, it shows 0 in the scriptlines.

Moreover, this patch ensures that the script of button "PendingMessagesRow Behavior" does compile in strict compilation mode.

Diffs:
In script of `button  "PendingMessagesRow Behavior" of card "templates" of stack "Message Box"`:

screen shot 2017-02-28 at 16 32 10


livecodeali:
@livecode-vulcan review ok 67e1409

livecode-vulcan:
💙 review by @livecodeali ok 67e1409bfd808229c0a7f47ee2ae975ceb34a0e3

PR #1560, opened 2017-03-12
[Bug 19410] Ensure the grid spacing can be set from Preferences

livecodepanos:
Changes in `scrollbar "grid" of group "Appearance" of card "Appearance" of stack "revpreferencesgui.rev"`:

screen shot 2017-03-12 at 18 23 15

It seems that for some reason from LC 8.1.0, the `the endValue of me - x` is not always an integer (i.e. 5.999879 instead of 6), and the "gridSize" field does not like that. So, this patch uses `round()` to ensure this difference is always an integer.


livecodepanos:
This patch is a dirty workaround. The fact that `the endValue of me - x` is not an integer is a bug in the engine, so the appropriate fix should be applied in the engine code, not the IDE.

Thus I am closing this PR.

PR #1565, opened 2017-03-29
[[ Tests ]] Fix sqlite standalone inclusion test

runrevmark:
This patch corrects the string to check for in the event of the
database driver not being found - it should be 'revdberr' and not
'dberr'.


livecodeali:
@livecode-vulcan review ok d4004e6

livecode-vulcan:
💙 review by @livecodeali ok d4004e671d051ec465d00fdde18783a9f2d1573b

PR #1577, opened 2017-04-18
Load diff library into IDE session at startup

peter-b:


runrevmark:
@livecode-vulcan review ok ecd4818

livecode-vulcan:
💙 review by @runrevmark ok ecd481893e6ddb79d5207fffb4ecaea8a77b4b46

PR #1581, opened 2017-05-02
[[ Bug 19564 ]] Defer deletion of script tab

livecodeali:
The script tab group gets deleted in a mouseUp handler of one of
its children, which causes an exec error. This patch defers the
deletion to after the mouseUp has been handled, using `send ...
in 0 millisecs`


livecodepanos:
@livecode-vulcan review ok c3480b1

livecode-vulcan:
💙 review by @livecodepanos ok c3480b1cadb0aa86c2bad2de7fd0935372f60678

PR #1584, opened 2017-05-04
revtemplatechooser is no longer used

mwieder:
The revtemplatechooser.rev stack appears to be an artifact from the revMedia days. It's not referenced from anywhere and the buttons all have code that explicitly does nothing. While it would be nice someday to have this sort of functionality as templates for new users (especially pluggable/maintainable), today the samples seem to serve the same purpose.


montegoulding:
@mwieder do you think you could `git commit --amend` to prefix the summary line with `[[ Cleanup ]] ` then add an empty line followed by your PR summary message? If you force push after the amend it will update the PR.

mwieder:
Done. Didn't know [[ Cleanup ]] was one of the valid tags. Er... did you want the whole github pr comment as the commit comment?

montegoulding:
Yeah... should be:

```
[[ Cleanup ]] revtemplatechooser is no longer used

The revtemplatechooser.rev stack appears to be an artifact from the revMedia days. It's not 
referenced from anywhere and the buttons all have code that explicitly does nothing. While it 
would be nice someday to have this sort of functionality as templates for new users (especially
pluggable/maintainable), today the samples seem to serve the same purpose.
```

livecodeali:
@livecode-vulcan review ok 31d263c

livecode-vulcan:
💙 review by @livecodeali ok 31d263c430a86881912d117321b6658f83df697f

PR #1573, opened 2017-04-06
[[ Bug 19264 ]] Ensure LCB errors display reasonably in SE

livecodeali:

PR #1586, opened 2017-05-05
[[ Bug 19264 ]] Clean up message box error display

livecodeali:
https://github.com/livecode/livecode-ide/pull/1573 only partially fixed the display of LCB errors - this patch ensures the whole error is passed through to the IDE library to format the error correctly when it is thrown from execution in the message box.

It also fixes an issue with external handler errors from the message box, that were previously not being handled.


montegoulding:
@livecode-vulcan review ok 68e60aa

livecode-vulcan:
💙 review by @montegoulding ok 68e60aa82856f9dc3ad644260cd9fcba9c68a8b6

PR #1587, opened 2017-05-12
[[ Bug 19480 ]] Overhaul message box execution

livecodeali:
This patch simplifies message box execution by relying exclusively
on compile checking to find the correct form of the script to
execute, and always using the `send <script blob> to obj` form to
execute (in non-debug mode).

The message box has been changed to always display the (single)
context object, either the active stack, selectedObject: <obj>, or
mouseControl <obj>. Thus it is always explicit what the context of
message box execution is.

This also fixes bug 19480 since the latter was caused by the script
being executed a second time after throwing an execution error in
the context of the intelligence object.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -11354,10 +11354,22 @@ private command __removeMessageBoxFromDebugContexts @xContexts
    delete line tFirst to tLast of xContexts
 end __removeMessageBoxFromDebugContexts
 
+function ideDeclaredGlobalVariables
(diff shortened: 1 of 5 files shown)


livecodeali:
screen shot 2017-05-12 at 21 51 30
screen shot 2017-05-12 at 21 51 45
screen shot 2017-05-12 at 21 52 15

montegoulding:
@livecode-vulcan review ok 385b4c3

livecode-vulcan:
💙 review by @montegoulding ok 385b4c3a9c9c77d5230f1a3f552d9eec8365df42

PR #1597, opened 2017-05-26
[[ LegacyInks ]] Remove legacy ink use from IDE

montegoulding:
This patch removes the use of legacy inks from the IDE and replaces
them with appropriate supported inks. It also deletes revupdatechecker
which is not used and very out of date.

Note the patch was scripted so should it cause merge conflicts when merging up to develop we can just run it again.


livecodeali:
@livecode-vulcan review ok 7831b50

livecode-vulcan:
💙 review by @livecodeali ok 7831b5028873cdea31dc2eb34d6cd0c2051aa9e6

PR #1611, opened 2017-06-08
[[ Bug 19824 ]] 'is scripted' step is never 'is not'

livecodeali:
Pass false as the 'is not' parameter to the check that a property
has a given value when the step is an 'is scripted' step - this is
never an 'is not' situation.


runrevmark:
@livecode-vulcan review ok 474f50d

livecode-vulcan:
💙 review by @runrevmark ok 474f50d13a3f62c1aec299f0c79a981bfbda71c9

PR #1623, opened 2017-06-09
[Bug 19839] Declare vars in scripts needed for BMI tutorial

livecodepanos:


livecodepanos:
@livecode-vulcan review by @runrevmark ok 4e5ab92

livecode-vulcan:
💙 review by @runrevmark ok 4e5ab924d5358ad24f16b0f912a9a7ebf14573e1

runrevmark:
Oops - I missed those changes - I'm not sure this is appropriate for the tutorial - we don't want new users encountering explicitVariables in the first instance... This suggests that the tutorial itself needs to toggle Script Compilation Mode.

We should try and fix this properly on Monday.

livecodeali:
The tutorial *does* toggle strict compilation... not sure what's going on here :-(

livecodepanos:
Hmm maybe it was not unchecked in my case because I was running from Xcode, where `grevdevelopment` is true and in this case you cannot uncheck strict compilation (it is re-checked automatically)

livecodepanos:
off-topic, happy birthday @livecodeali :)

PR #1626, opened 2017-06-12
[[ Tutorial ]] Don't show 'Do It For Me' in a popup answer dialog step

livecodeali:
If a tutorial step is `wait until button <name> pops up answer dialog`, the skipper doesn't know what to pop up or how, as it could be triggered from script when clicking the button (as it is here), or some other way.

So, don't show the option to 'Do It For Me' in this case.


runrevmark:
@livecode-vulcan review ok 17964ae

livecode-vulcan:
💙 review by @runrevmark ok 17964ae8f0ddc46a74146f0baf1e5edb1ac98f96

PR #1627, opened 2017-06-12
[[ Bug 19836 ]] Add stack rect step to Todo list tutorial

livecodeali:
Wait actions for creating stacks don't know about the IDE stack size
choices, so add a step for resizing the stack in case the user
chooses a different size, or decides to skip the step.

The step will be automatically skipped if the user chooses the
indicated size from the menu.


livecodepanos:
@livecode-vulcan review ok d161cb5

livecode-vulcan:
💙 review by @livecodepanos ok d161cb55cf571cd522fc4140a050ed75e7cd217f

PR #1631, opened 2017-06-14
[Bug 19347] Allow setting the 'style' of the stack from the PI

livecodepanos:


livecodepanos:
This PR fixes the problem, but I am not sure if setting the `style` of the stack should be feasible from the Property Inspector. If someone sets the style to `modal` *accidentally*, there is no way to undo that, other than force-quitting the IDE.

montegoulding:
@livecodepanos what about checking if the stack is currently open before setting the style to modal and throwing up a warning dialog first if it is? We have the same issue from the contextual menu I think so perhaps this can be handled centrally.

Toolset/resources/supporting_files/property_definitions/com.livecode.interface.classic.Stack.txt
```diff
@@ -7,7 +7,7 @@ name
 title	
 mainStack		
 decorations												
-style							Toplevel,Modeless,Modal,Palette
+style							Toplevel	Toplevel,Modeless,Modal,Palette
```
livecodeali, line 10: If you do `style::revIDESetModeOfStack` here, it will use the `revIDESetModeOfStack` as the property setter, so you can remove your change to `revIDEPropertySet` (in theory)

PR #1637, opened 2017-06-23
[[ Bug 19890 ]] Remove the possibility of bringing up palettes for incorrect objects

livecodeali:
- Ensure object is always selected when step requires
If an object waiting for a palette to open becomes deselected, go
back to the selection step.
Also, ensure this works for stacks by implementing similar code in
ideActiveStacksChanged to ideSelectedObjectChanged

- Ensure all palette for obj steps are preceded by select steps
This is required for the enforcement of correct selection to work.
Also, ensure all select steps highlight the object in question.


montegoulding:
@livecode-vulcan review ok b1b9394

@livecodeali this looks fine but I can't help wondering if we should just have something like this that you can add to any step.
```
ensure <object> is selected [with message <info>]
```

livecode-vulcan:
💙 review by @montegoulding ok b1b93947a930dcb4965bb4a2730edea5fc19d047

PR #1638, opened 2017-06-23
[[ Tutorial ]] Allow instruction window to be moved

livecodeali:
- Turn pointer 'object' into array
The 'pointer' that is passed to various utility functions is now
an array, with its old value (the string top-left, top-right etc)
now under the array key "position".

- Tweak pointer API for top left override
Add to pointer API so that the topleft of the content rect of the
tutorial stack can be overridden

- Allow tutorial window to be dragged 
This patch uses the modified pointer API to allow the tutorial
message stack to be dragged anywhere on the screen while continuing
to point to the intended target.

### DIFF
--- a/Toolset/palettes/tutorial/revtutorial.livecodescript
+++ b/Toolset/palettes/tutorial/revtutorial.livecodescript
@@ -1103,10 +1103,6 @@ command revTutorialPositionStackAndPointer pContentWidth, pContentHeight, pHighl
 end revTutorialPositionStackAndPointer   
 
 command revTutorialCreateStack pContentWidth, pContentHeight, pPointer
-   if the platform is "win32" then
-      set the windowshape of stack "revTutorial" to empty
-   end if
-   
    # Store templategraphic props (in case we have a graphic tool selected)
    local tProps
    put the properties of the templategraphic into tProps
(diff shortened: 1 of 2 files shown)


livecodeali:
Warning - this really needs to be tested on Windows.

Toolset/palettes/tutorial/revtutorial.livecodescript
```diff
@@ -482,20 +482,20 @@ on revTutorialParseMenuHighlight pTokens, pStart, @rData
...
-      revTutorialParseSubexpression tExpressionTry, pTokens, pStart, rData
-      if the result is empty then
-         return empty
-      end if
-      put tSubMenuString & " " after tSubMenuExpressionTry
-      add 1 to tSubMenuTry
-   end repeat
-   
-   return "Expected: highlight [menu item <token> of...] menu <token>"
+   put "highlight " into tExpressionTry
```
montegoulding: @livecodeali what's happened to the indenting here? Is this a script editor issue?

livecodeali: oh, weird...not sure. Will fix

livecodeali: looks like it is an SE bug, will see if I can reproduce

montegoulding:
@livecode-vulcan review ok 3273dfd

livecode-vulcan:
💙 review by @montegoulding ok 3273dfd4faf488ea0a9c5d6f91d33f1f215e8bdb

PR #1643, opened 2017-06-27
[[ Bug 19961 ]] Fix set property editor

livecodeali:
The height of editor components needs to be dependent on the content
rather than the height of the group containing the content, otherwise
it is not fixed, which causes the layout to be calculated incorrectly.


montegoulding:
@livecode-vulcan review ok 1b4b8bb

livecode-vulcan:
💙 review by @montegoulding ok 1b4b8bb7953f53e812911cc8c31dc2c776d43d94

PR #1652, opened 2017-07-03
Fixed Some Labels on MacOS

bhall2001:
On Mac OS, some of the field labels and 1 menu selection where truncated when the text size was increased to 12pt size. This fixes the following items:

Appearance Card
  - Fixed “Documentation Hypertext” Label
  - Moved settings to right for label to fit
  - Adjusted width of Documentation Hypertext menu to show all text
Object Size Card
  - Fixed “Progress bar (MacOS):” label
Script Editor Card
  - Fixed “When editing scripts:” label
Updates Card
  - Updated “Check For Updates” button height to 23
Project Browser Card
  - Updated “Object type indicator:” label
  - Moved settings to the right for label to fit
  - Better alignment of label to radio buttons and checkboxes


montegoulding:
Hi @bhall2001 unfortunately it's difficult for us to accept contributions to binary stacks. revPreferencesGUI is particularly difficult because there are different versions in version 8.1 and 9 already and changes have to me manually merged. If you could turn this patch into a script we could execute to set the object properties then we could apply it.

bhall2001:
Will do. I was wondering about that myself. I’ll put together a script that makes the changes.

Should I eMail the script to you or post it in the pull request?

Thanks and sorry about that,
Bob

montegoulding:
Don't be sorry @bhall2001 we really appreciate the contribution. It's just hard with binary stacks. How about I close this and you create a bug report when you can including the script that fixes the field sizes. Everything is meant to have a bug report anyway so that we can keep track of things.

bhall2001:
Will do.

PR #1666, opened 2017-07-12
[[ Default Handlers ]] Various fixes

livecodeali:
- [x] Prevent removal of P from handler name
- [x] Make default handler name grey in list
- [x] Fix spacing of default handler list
- [x] Add SE preference for showing list

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsehandlerlistbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsehandlerlistbehavior.livecodescript
@@ -140,7 +140,9 @@ command update
    put the result into tHandlers
    
    local tDefaultHandlers, tIndex
-   put revSEDefaultHandlers() into tDefaultHandlers
+   if sePrefGet("editor,showDefaultHandlers") then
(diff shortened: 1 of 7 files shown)


montegoulding:
@livecodeali I have added a note to 20066 to Hungarianize the params. Particularly important as some params will be reserved words.

livecodeali:
The existing ones are dealt with here: https://github.com/livecode/livecode/pull/5693

But I agree, no reason not to make absolutely sure wherever possible! Particularly as we will not have control over the widgets' ones.

montegoulding:
Hmm... it is probably best to deal with it via the docs so they are the same in both places. We should add a test to ensure that messages have hungarianized parameter names.

livecodeali:
screen shot 2017-07-15 at 15 53 16
screen shot 2017-07-15 at 15 53 26
screen shot 2017-07-15 at 15 53 37

runrevmark:
@livecode-vulcan review ok 532246b

livecode-vulcan:
💙 review by @runrevmark ok 532246b0753a865cda16fc87816456d5267a97e6

PR #1668, opened 2017-07-14
[[ Bug 20041 ]] Update tags & wait array when a stack name changes

livecodeali:
Update the long IDs in the tagged object array for the new stack name, and update the wait array so that we re-subscribe to the messages we need with any updated long IDs


runrevmark:
@livecode-vulcan review ok 585ccc8

livecode-vulcan:
💙 review by @runrevmark ok 585ccc80e79856f3831728f116621fb4613fef16

PR #1670, opened 2017-07-15
[[ Tutorial ]] Use smaller images in To Do List tutorial

livecodeali:
- Add mechanism to include smaller version of images if they are too tall for a smaller screen
- Add smaller versions of To Do List images


runrevmark:
Would it be possible to make 'small' image generation automatic - i.e. by downsizing the normal size images if a certain screen size criteria is not met?

livecodepanos:
@livecode-vulcan review ok 5790c72

livecode-vulcan:
💙 review by @livecodepanos ok 5790c72b6042fa6ec9f83b5edf2b40dd1174aea9

livecodeali:
That might make more sense, although then you have to add logic for the given image whether it actually needs resizing for smaller screens or not. On balance probably better than adding more images to the repo

runrevmark:
I've added http://quality.livecode.com/show_bug.cgi?id=20101 to track this.

@livecode-vulcan review ok 5790c72

PR #1682, opened 2017-07-18
[Bug 18366] Enable control-tab on Mac OS

bhall2001:
Allow use of control-tab & control-shift-tab on Mac OS to cycle between tabs in the Script Editor.

### DIFF
--- a/Documentation/guides/The LiveCode IDE.md
+++ b/Documentation/guides/The LiveCode IDE.md
@@ -784,6 +784,8 @@ Figure 12 – Find and Replace
 | Find next                                      | Control-g                               | Command-g                               |
 | Find selected text                             | Control-Option-f                        | Command-Option-f                        |
 | Format current handler                         | Tab                                     | Tab                                     |
+| Next tab in Editor                             | Control-tab                             | Control-tab                             |
+| Previous tab in Editor                         | Control-shift-tab                       | Control-shift-tab                       |
 
 | **The Message Box** |  **Windows / Linux** | **Mac OS X**|
 |------|--------|-------|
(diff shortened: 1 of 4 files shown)


livecodeali:
@bhall2001 Thanks for this! Is there any reason why it's restricted to MacOS?
There is a table in the docs for keyboard shortcuts relating to the script editor (https://github.com/livecode/livecode-ide/blob/develop/Documentation/guides/The%20LiveCode%20IDE.md#keyboard-shortcuts) so this should be added to that.

mwieder:
@livecodeali not necessarily restricted to MacOS: control-tab and shift-control-tab are already supported for cycling through script editor tabs on linux, so the patch isn't necessary there. I'm too lazy to test Windows.

livecodeali:
Ah makes sense, I had misread the patch slightly. In that case, it just needs adding to the shortcuts list in the guide.

livecodeali:
@bhall2001 As far as I'm concerned this is fine to merge once the shortcut has been added to the shortcuts table in the guides

bhall2001:
I was surprised to see this shortcut was not already in the Docs. And to be clear, I have not changed any of the existing functionality for Linux and Windows. I replaced command-tab with control-tab on MacOS since command-tab is caught by the OS and never reaches LC.

livecodeali:
@bhall2001 Thanks! I'm a little confused about that last commit - why is it not Control + tab on Windows/Linux?

livecodeali:
@livecode-vulcan review ok 81cd8a8

livecode-vulcan:
💙 review by @livecodeali ok 81cd8a84df73c4cb57d1e5865078e636c3a61e71

PR #1683, opened 2017-07-18
[[ WIP ]] Scriptify IDE preferences gui

livecodeali:


livecodeali:
This sort of thing
screen shot 2017-07-18 at 22 22 56
screen shot 2017-07-18 at 22 23 00
screen shot 2017-07-18 at 22 23 18

montegoulding:
This is a great idea @livecodeali we should bump it up the priority list

mwieder:
+1 I agree, despite my snarky comments on the use-list.
And more so if there's an api to set/retrieve preferences through the ide.

bwmilby:
Probably a good time to ensure there are no prefs included that are no longer referenced/used.  I like the new look.

PR #1687, opened 2017-07-20
[[ Bug 18278 ]] Dictionary Whitespace

bwmilby:
Reorganized the HTML file and adjusted CSS/JS to eliminate the white space around the dictionary panels.  Uses percentages for column widths.  Tested with DP8 on Windows 10 and Mac OS X 10.12.5.  What is interesting is that on Windows, changes to the template/css/js are immediately viewable without restarting LiveCode but on the Mac I had to restart LiveCode to see the changes.

CSS:
Static vertical scroll bar added to prevent the header and filter panel from resizing every time the definition length changed. (not much impact on Mac)
Added ID for navbar.
Replaced resizer image to one that fits the narrow window (automatically centers).
Removed the affix code (changed to position:fixed in filter panel only).
Added scroll bar to dropdown menu (DP8 list was too long to display in the default size).

HTML:
Changed "container" to "container-fluid" to use the entire screen width.
Created a static navbar at the top of the window.  Moved the "Choose API" label to the bar.
Most of the deletes/adds are relocating elements into the navbar.
Removed "affix" from the list to allow for more room to see the definition.

JS:
Small tweak to container height calculation.
Removed "update_lcdoc_margin" function as it is no longer needed since the list is no longer fixed.
Moved "Choose API:" label from JS to HTML
Scrolls to top of window when typing in search box since list no longer fixed.

IMG:
Updated vertical resizer image needed since original copy was too wide to center on narrow pages


livecodeali:
@bwmilby Really superb work, thank you very much.
@livecode-vulcan review ok 117f10d

livecode-vulcan:
💙 review by @livecodeali ok 117f10d72b26386c4cad723f663db85cb7ad80de

PR #1692, opened 2017-07-22
[[Bugfix 15302]] IDE occured -> occurred

bwmilby:
Spelling correction of occured to occurred in IDE subproject.
errors.livecodescript is causing the build to fail on the main project
Also submitted pull request for marked.js with the originator of the code

Errors not corrected in initial pull request:
There are 6 errors in Toolset\palettes\revresourcecenter.rev
There is 1 error in Plugins\livecodeTestInterface.livecode but the script has a compile error

https://github.com/livecode/livecode/pull/5725 is the corresponding engine PR


tests/core/idelibrary/errors.livecodescript
```diff
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
 
 constant kExtensionErrorCode = 863
-constant kExpectedError = "extension: error occured with domain"
```
bwmilby, line 20: This change is dependent on executionerrors.h being modified as well.  It seems that this test was chosen as the first 8.0 specific error.  To avoid the "occured" to "occurred" changes in the .h, this test could be changed to 867 to make it independent of the updates to the .h file.

montegoulding:
@bwmilby unfortunately we won't be able to accept the patch if it has changes to binary stack files as it's hard for us to review and ensure changes are limited to those described in the commit message. If you think those changes are important then providing a script that we can run would work. It also means that in the event of a merge conflict we can run the script again. Personally I don't think the trouble of dealing with a patch to revCore.8.livecode is worth it just for an errant comment. This patch will also need to be rebased on develop-8.1 as the livecode one was.

bwmilby:
Understand.  I'll pull the binary files out and rebase.  Any thoughts on the update to the errors.livecodescript test?  If that file is fixed then builds will fail until the the executionerros.h file is also updated.  If the goal of the test is to make sure the lookup between code and message matches then it may be good to change it to look at the first message without the "occured" error (867).

Documentation/html_viewer/js/marked.js
```diff
@@ -2,4 +2,4 @@
...
...one(err);if(code==null||code===token.text){return--pending||done()}token.text=code;token.escaped=true;--pending||done()})})(tokens[i])}return}try{if(opt)opt=merge({},marked.defaults,opt);return Parser.parse(Lexer.lex(src,opt),opt)}catch(e){e.message+="\nPlease report this to https://github.com/chjj/marked.";if((opt||marked.defaults).silent){return"<p>An error occurred:</p><pre>"+escape(e.message+"",true)+"</pre>"}throw e}}marked.options=marked.setOptions=function(opt){merge(marked.defaults,opt);return marked};marked.defaults={gfm:true,tables:true,breaks:false,pedantic:false,sanitize:false,smartLists:false,silent:false,highlight:null,langPrefix:"lang-",smartypants:false,headerPrefix:"",renderer:new Renderer,xhtml:false};marked.Parser=Parser;marked.parser=Parser.parse;marked.Renderer=Renderer;marked.Lexer=Lexer;marked.lexer=Lexer.lex;marked.InlineLexer=InlineLexer;marked.inlineLexer=InlineLexer.output;marked.parse=marked;if(typeof module!=="undefined"&&typeof exports==="object"){module.exports=marked}else if(typeof define==="function"&&define.amd){define(function(){return marked})}else{this.marked=marked}}).call(function(){return this||(typeof window!=="undefined"?window:global)}());
```
livecodeali, line 5: This must be a similar error in the marked.js source! No harm fixing it here of course.

tests/core/idelibrary/errors.livecodescript
livecodeali: I suggest we just merge and temporarily break the branch

livecodeali:
@livecode-vulcan review ok 1a59bf1

livecode-vulcan:
💙 review by @livecodeali ok 1a59bf16325f6dc92ca12231d5f94676e93aa62e

PR #1693, opened 2017-07-23
[[ Bug 18881 ]] Sort API Dictionaries

bwmilby:
Add custom sort function to dictionary_functions.js and call upon load to sort the API dictionaries.
Keeps "LiveCode Script" and "LiveCode Builder" as the first two entries.
LC requests to display an entry use the dictionary name vice ID.

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -9,6 +9,11 @@
 	function dataGet(){
 		//console.log(dictionary_data.docs);
 		
+		//sort API dictionaries, leave LiveCode Script/LiveCode Builder as first 2 entries
+		if(tState.selected_api_id == ""){
+			dictionary_data.docs = dictionary_data.docs.sort(compareDictionaryObject);
+		}
+
 		if(!dictionary_data.docs.hasOwnProperty(tState.selected_api_id)){
 			
 			$.each(dictionary_data.docs, function(index, libraryData){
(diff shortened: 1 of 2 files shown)


Documentation/html_viewer/js/dictionary_functions.js
```diff
@@ -25,6 +30,29 @@
...
+			return -2;
+		if(entryObject2["display name"] == "LiveCode Builder")
+			return 2;
+		if(entryObject1["display name"].toLowerCase() < entryObject2["display name"].toLowerCase())
+			return -1;
+		if (entryObject1["display name"].toLowerCase() > entryObject2["display name"].toLowerCase())
+			return 1;
+		return 0;
+	}
+
```
livecodeali, line 50: You don't need that many comparisons for LCS and LCB, it would be slightly more efficient as
```
		if(entryObject1["display name"] == "LiveCode Script"
			return -1;
		if(entryObject2["display name"] == "LiveCode Script"
			return 1;
		if(entryObject1["display name"] == "LiveCode Builder")
			return -1;
		if(entryObject2["display name"] == "LiveCode Builder")
			return 1;
...
```

bwmilby: Looks like you are correct.  I thought that I tried that combination when testing before I added the 2 & 3.  I'll verify this evening on my Mac and update the PR.

livecodeali:
@livecode-vulcan review ok c4aa0ab

livecode-vulcan:
💙 review by @livecodeali ok c4aa0abc2b50913cb7397086249eaff95cb949d5

bwmilby:
I see that "LiveCode IDE" has now been added in the develop branch.  To make it third will require a couple additional lines in the sort code.

livecodeali:
@bwmilby that is only generated when building from source at the moment. In develop, we will probably roll all the loaded extension APIs into the LiveCode Script API to make things more discoverable (for example at the moment you have to know what extension a piece of syntax is implemented in to find it). Anyway, i say all this not to prevent anyone from making the change if they'd like, but just FYI 🙂

PR #1700, opened 2017-07-27
[[ Bug 7303 ]] Implement autocomplete in script editor

montegoulding:
This patch adds autocomplete to the script editor. Autocomplete providers
are generated by object introspection using revAvailableHandlers and
revAvailableVariables. Autocomplete providers are also generated from
the syntax statements in the documentation database.

PR #1701, opened 2017-07-31
[[ Bug 7303 ]] Cherry-picked auto-complete on develop-8.1

montegoulding:
I will leave PR #1700 open while we work out which version autocomplete should go in. This patch should be everything required for autocomplete and bracket completion.

### DIFF
--- a/Documentation/guides/The LiveCode IDE.md
+++ b/Documentation/guides/The LiveCode IDE.md
@@ -565,6 +565,22 @@ Choose a handler to go to that handler in the script.
 
 The Window menu contains the names of open script editor windows.
 
+### Autocomplete
+
+While typing in the script editor a filtered list of available completions
+will appear below the selection. Use the following keys to navigate the
+completion list:
+
+- tab - apply the completion
+- up arrow - move up the list
+- down arrow - move down the list
+- right arrow - move to the next placeholder field
+- left arrow - close the completion list
+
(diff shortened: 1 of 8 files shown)


livecodeali:
This is definitely good enough to go as far as I can see - hopefully any minor issues can be ironed out in the release process (although I don't see any yet)
@livecode-vulcan review ok 677c401

livecode-vulcan:
💙 review by @livecodeali ok 677c4013ca1854b282aa506a87b048d61a68de6d

livecodeali:
(obviously this shouldn't be merged before https://github.com/livecode/livecode/pull/5746)

montegoulding:
@livecodeali I've found a couple of things I'd like to tweak while using this today

- I often type something quickly then hit tab to format the handler and this is catching the autocomplete and messing up what I typed which is rather annoying (haven't looked into this yet)
- The left and right arrow key handling appears to be both unnecessary and it occasionally breaks moving the cursor so that needs to be removed

montegoulding:
@livecodeali I think I've sorted out the above issues and the issue where occasionally the completions weren't showing up

livecodeali:
@livecode-vulcan review ok cdaad06

livecode-vulcan:
💙 review by @livecodeali ok cdaad065869682bc7b373ea2a17af5df106b7a44

PR #1709, opened 2017-08-11
[[ Bug 7303 ]] Implement placeholder & autocomplete support in SE

montegoulding:

PR #1710, opened 2017-08-13
[[ Bug 20258 ]] Improve control structure completion

montegoulding:
This patch adds control structure completion for `switch`
`case` to be completed with `break`. This change will reduce
continuation errors in user scripts.

This patch also adds control structure completion for `if` when
the last token of the line is `then`. Due to the numerous variants
of `if` this was the only variant with a low risk of regressions.

Note that this is a separate PR But uses the __GetPreference API added in #1709 so should be merged after that if that get merged.


livecodeali:
@livecode-vulcan review ok f9f9cf0

livecode-vulcan:
💙 review by @livecodeali ok f9f9cf0905a10a04f7215f0390915b506ffecddf

PR #1713, opened 2017-08-28
[[Bug 20289]] Improve dictionary sort

bwmilby:
Initial complaint was that "me" is very difficult to locate in the dictionary.  The current priority search ends up with terms in an unsorted state since it does not explicitly maintain the alphabetical sort.  Additionally, the prioritized sort is looking at all syntax terms which is probably more than is needed for the priority portion.

Added a new regex for an exact word match.  This will be the first one used for the sort and will use the display name.  This will allow for things like "me" to appear at the top and other short words to be closer to the top.  The second regex will use the existing "start of word" search, but also only search the display name.  Finally, the display names will be compared to maintain alphabetical order.

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -85,20 +85,24 @@
    		var tokensOfTerm = tTerm.match(/\S+/g);
 		
 		
-		// Generate two regexes - one that matches all syntax that 
+		// Generate three regexes - one that matches all syntax that 
 		// contains each search term, and one that matches all syntax that
 		// contains a word beginning with each search term. This way we 
-		// can prioritise matches that start with the search terms.
-    	var matchExp = '';
-    	var priorityMatch = '';
-    	$.each(tokensOfTerm, function(index, matchToken)
-    	{
-       		matchExp += '(?=.*' + matchToken + ')';
-       		priorityMatch += '(?=.*\\b' + matchToken + ')';
-   		});
+		// can prioritize matches that start with the search terms.
+        // Third regex will match full words.
+        var matchExp = '';
+        var priorityMatch = '';
+        var wordMatch = '';
+        $.each(tokensOfTerm, function(index, matchToken)
+        {
+            matchExp += '(?=.*' + matchToken + ')';
+            priorityMatch += '(?=.*\\b' + matchToken + ')';
+            wordMatch += '(?=.*\\b' + matchToken + '\\b)';
+        });
 
-		var regex = new RegExp(matchExp, "i");
-		var priorityRegex = new RegExp(priorityMatch, "i");
+        var regex = new RegExp(matchExp, "i");
+        var priorityRegex = new RegExp(priorityMatch, "i");
+        var wordRegex = new RegExp(wordMatch, "i");
 		
 		// Grep for the general search term
 		tState . searched . data = $.grep(tState.filtered_data, function (e) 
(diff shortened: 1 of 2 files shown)


Documentation/html_viewer/js/dictionary_functions.js
```diff
@@ -85,20 +85,24 @@
...
+		// can prioritize matches that start with the search terms.
+        // Third regex will match full words.
+        var matchExp = '';
+        var priorityMatch = '';
+        var wordMatch = '';
+        $.each(tokensOfTerm, function(index, matchToken)
+        {
+            matchExp += '(?=.*' + matchToken + ')';
+            priorityMatch += '(?=.*\\b' + matchToken + ')';
+            wordMatch += '(?=.*\\b' + matchToken + '\\b)';
```
montegoulding: @bwmilby @livecodeali I'm wondering if this should be tweaked so that the sort order is:

1. whole match of display name
2. display name begins with token
3. display name contains token
4. syntax contains token somewhere

bwmilby: That would be a very easy thing to add.  I'm not sure if we should add 3 or replace 2 with 3.  I'm also wondering how this is impacting the speed of the sort.  It seems faster, but I want to do some actual tests and compare the original to the update (could even do a/b/c with above as c).  I need to figure out the best way to do that - probably with a stand alone browser.

livecodeali: I think we might need to keep 3. because of LiveCode Builder's habit of having display names that are distinct from the syntax.

Documentation/html_viewer/js/dictionary_functions.js
```diff
@@ -108,19 +112,35 @@
...
+                return 1;
+
+            // display names with words that start with terms prioritized
+            var tPriorityA = priorityRegex.test(tNameA);
+            var tPriorityB = priorityRegex.test(tNameB);
+            if (tPriorityA && !tPriorityB)
+                return -1;
+            if (tPriorityB && !tPriorityA)
+                return 1;
+
```
bwmilby: This is where the addition would be made to include 3. above.  I would probably want to change to using tMatchA & tMatchB for each of the 3 sets of Regex.  The third would use regex.test(tName?).  This will slow things down some, but I have not had a chance to build a page with timers to check all of the performance variations.

montegoulding: @bwmilby The alphabetical sort would mean that wordRegex isn't necessary wouldn't it or is it faster to return that here?

bwmilby: wordRegex will ensure an exact match that isn't the first word appears at the top of the list.

bwmilby: The wordRegex groups whole word matches at the top of the list so "pane" will have the 2 entries that match ahead of the ones with "panel" for example.  Without wordRegex, they are not separated and are just sorted alphabetically.

I'm trying to envision where adding the extra regex will help find things.  "the" is an example where it would change the order, but I'm not sure if it is helpful.  After "there" matches, you either get "dither" or "abs" as your next match.  Adding the extra regex will put matches in the middle/end of a word in the display name ahead of ones that have that kind of match only in the syntax.  I have the change stashed and can submit.

montegoulding: sorry... I was being stupid and thinking the word and priority matches were for the beginning of the display name not anywhere in it

livecodeali:
@livecode-vulcan review ok ebe217a

livecode-vulcan:
💙 review by @livecodeali ok ebe217a7bf74638927ea3e330428091ca01012fe

PR #1719, opened 2017-09-06
[Bug 20330] Do not allow stack names to contain quotes

livecodepanos:


montegoulding:
@livecode-vulcan review ok 92f6976

@livecodepanos I wonder if this should be at engine level for _any_ object name. Also comma would be a good thing to check for.

livecode-vulcan:
💙 review by @montegoulding ok 92f6976f27c0ddfff007c3cc6addd7c92eb86ee2

montegoulding:
@livecodepanos actually comma is handled at engine level with:

```
MCStringFindAndReplaceChar(*t_new_string, ',', '_', kMCCompareExact);
```

... perhaps there should be a general `sanitize_object_name` that ensures chars that we really don't want in names are cleaned out

PR #1721, opened 2017-09-07
[[ Bug 20342 ]] Save script selection when updated

montegoulding:
This patch ensures that `contextSave` and therefore the script selection
is called at times when the script is `the focusedObject` allowing the
selection to be restored when the user returns to the tab or re-opens
the script.


livecodeali:
@livecode-vulcan review ok 8be2c59

livecode-vulcan:
💙 review by @livecodeali ok 8be2c59de100528bde697f0efc2e7885b6f52627

PR #1725, opened 2017-09-08
[[ Bug 20264 ]] Fix infinite loop

montegoulding:
In the case where a line was not continued and the right bracket
was the first char of the line then the script editor would not
reach the end condition when looking for a bracket pair and remain
in the infinite loop.


runrevmark:
@livecode-vulcan review ok 2b98390

livecode-vulcan:
💙 review by @runrevmark ok 2b983900eb22fc7ff8840c102c1fd757f454ef42

PR #1727, opened 2017-09-08
[[ CommunityUpgrade ]] Add start center browser and menu button

livecodeali:
This patch adds a browser to the start center and a button to
the menu bar both of which, when clicked, open a browser
displaying the upgrade options in a browser widget over the
start center.


montegoulding:
I'm not sure I understand why it's necessary to have the extra edition param. Why not return a url from the edition properties that is for upgrading from it? The reason I'm being a bit nit-picky here is because the point of the api is to not have edition wrangling elsewhere.

livecodeali:
They also need to get the upgrade-edition's colour for the upgrade button on the menubar. But that could also be an extra item in the switch if you want? I just didn't see the need for the proliferation of switch cases.

As a compromise, how about:
```
revEnvironmentEditionProperty pProp, pForUpgrade
local tEdition
put the editionType into tEdition
if pForUpgrade then
   put revEnvironmentEditionProperty("upgrade_edition") into tEdition
end if

switch tEdition
...
```

montegoulding:
Hmm... I guess the other question is does your current solution offer a level of flexibility we might need... we might feasibly need to get different edition colors and icons for the docs if we want to pretty up the edition element...

montegoulding:
It probably would be better to use the edition name as the param rather than the editionType which is really the license class. We could feasibly end up with multiple editions with the same license class. For example an education skinned community plus.

livecodeali:
There should perhaps be some way of fetching the actual edition name although I'm not sure there's an easy way of doing that...

montegoulding:
Do you mean engine level? I think edition name is really an IDE thing. Eventually the loaded edition can come from the edition.txt file with a license class sanity check rather just from the license class. Although if we never have a new edition without a new license class then perhaps that's not necessary ;-)

livecodeali:
I didn't mean engine level - sounds easier that I was thinking for some reason.

montegoulding:
The environment stack reads that file. It perhaps if it initialised a global variable with edition name?

montegoulding:
OK So I think we agreed to merge this as is and look at the API later if it's an issue

@livecode-vulcan review ok 489eeb2

livecode-vulcan:
💙 review by @montegoulding ok 489eeb28b5860c55affef96966b3b0e9d6b54a5f

PR #1729, opened 2017-09-10
[[ Tests ]] Improve test output

montegoulding:
I'm not sure if we actually want this if it doesn't tell us if the IDE crashed so marking WIP for now. It solves the IDE tests have no output until complete issue with Travis timing out. It also adds colour highlighting to the output.

### DIFF
--- a/tests/_testrunner.livecodescript
+++ b/tests/_testrunner.livecodescript
@@ -19,6 +19,8 @@ along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
 -- FIXME provide this on the command line
 constant kLogFilename = "_test_suite.log"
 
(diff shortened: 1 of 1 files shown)


montegoulding:
It appears that IDE tests can now take longer than 10 minutes on `develop` therefore I'm taking the WIP label off. Will need some conflict resolution when merging up to `develop`. We should prioritise this enhancement so we can reinstate process crash tests on the IDE:
http://quality.livecode.com/show_bug.cgi?id=19896

livecodeali:
@livecode-vulcan review ok 054d045

livecode-vulcan:
💙 review by @livecodeali ok 054d0458233ab40d038609ce85861e34a67e0121

PR #1730, opened 2017-09-10
[[ Bug 18739 ]] Dictionary auto-search on first char freezing cursor

bwmilby:
Did some preliminary timing at various points of the js.  The actual search `tData = dataSearch(pTerm);` was not the primary driver of the amount of time taken between typing a character and presentation of the information.  The search was in the 100ms range but the HTML generation `$("#table_list").html(tHTML);` would get up to 500ms frequently.  After researching methods to improve performance, the idea of delaying the actual search stood out as a preferred method for this use case.

Implemented a function (debounce) so that the search is not performed until 250ms after the last character is typed.  This would add 250ms to a single character search, but will more than overcome that when multiple characters are typed in quick succession.  If looking for something like "#", then the extra time is barely noticeable.  If typing multiple letters (especially 3 or more), the performance improvement is very evident.

The code is based on the following post:
https://stackoverflow.com/questions/12538344/asynchronous-keyup-events-how-to-short-circuit-sequential-keyup-events-for-speed
The actual lines of code are cut down from a function included in this project:
https://github.com/jashkenas/underscore
The project code is under the MIT License

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -1158,13 +1158,30 @@
 		return (typeof liveCode !== 'undefined');
 	}
 	
+	// Returns a function, that, as long as it continues to be invoked, will not
+	// be triggered. The function will be called after it stops being called for
+	// N milliseconds.  Used to make search more responsive.
+	var debounce = function(func, wait) {
+		var timeout;
+		return function() {
+			var context = this, args = arguments;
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok a28c198

livecode-vulcan:
💙 review by @livecodeali ok a28c198174664ece0f82bc6bbbcdaa2446b7e29f

PR #1732, opened 2017-09-11
[[ Live Errors ]] Improve live errors for explicit vars

montegoulding:
This patch improves live errors when the user has `the explictVariables`
preference `true`. Previously the compilation done to generate the script
description was done with `explicitVariables` set the `false`. This patch
changes that to attempt to compile with the user's preferred setting and
if that fails and the `explicitVariables` was `true` to try again without
it in order to extract the script description while retaining the errors
from the user setting. This won't be necessary if the compiler is changed
to allow a script description to be generated even if the compilation had
errors.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
@@ -517,13 +517,7 @@ command getHandlerList
 end getHandlerList
 
 private command updateHandlerList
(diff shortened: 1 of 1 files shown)


livecodeali:
@livecode-vulcan review ok 5d3acdd

livecode-vulcan:
💙 review by @livecodeali ok 5d3acdd0a43a159e926d165c00b7ad284dd376e3

PR #1735, opened 2017-09-13
[[ Bug 20374 ]] Fix deselecting of current line when adding handler

montegoulding:
This patch adds a select `after` mode to the `goLine` command and
uses that for slecting the line in the middle of the new handler
when the handler is added to the script editor. It also adds an
indent so the cursor is positioned correctly.


livecodeali:
@livecode-vulcan review ok 739dfc0

livecode-vulcan:
💙 review by @livecodeali ok 739dfc0e851408ea1b0f6b02bf8194fff0cca24b

PR #1739, opened 2017-09-14
[[ Bug 20386 ]] Options menu for some SE prefs

montegoulding:
We could feasibly add some more/all of the preferences here, however, I'm just looking at this as an interim measure until we refactor prefs. When we do so we might include a way to generate menus also I guess.


livecodeali:
@livecode-vulcan review ok 14f4657

livecode-vulcan:
💙 review by @livecodeali ok 14f465728ecfb784a0f44b4a56f405a46541bd7d

PR #1748, opened 2017-09-19
[[ Bug ]] Fix selection update issues

montegoulding:


livecodeali:
What's going on here then?

montegoulding:
Ah well there's two things:
- for some reason I added a check to see if selectionUpdateRequest was called with a target of the script field (not necessary and interferes with action based edits from the menu etc)
- selectionUpdateRequest is called in textReplace now so no need to also call it after these actions

montegoulding:
Sorry I can amend the commit msg with ^ if you like

livecodeali:
Yeah, I think that would be good. Looks fine though!

montegoulding:
done

livecodeali:
@livecode-vulcan review ok ef926dd

livecode-vulcan:
💙 review by @livecodeali ok ef926dda29a4d540958be6cf43663e36102a63cc

PR #1752, opened 2017-09-20
[[ Bug 20392 ]] Eat closing brackets when typing

montegoulding:
Because we have bracket completion and people are used to typing
closing brackets it saves time if when typing a closing bracket
that we look ahead and if the next char is already that closing
bracket then we just replace it rather than inserting another.
This means the selection will now be outside the brackets.


livecodeali:
@livecode-vulcan review ok 4a8b51a

livecode-vulcan:
💙 review by @livecodeali ok 4a8b51ae12f80746b05d40e9b2c26b451ad14143

PR #1760, opened 2017-09-22
[[ Bug 20444 ]] Use Source Code Pro as default SE font

montegoulding:
This patch adds the ttf files for Source Code Pro to the IDE supporting
files and starts using them when the IDE library is initialized. If the
user has not set a script editor font or chooses Default then the
preference will be empty. The two handlers that get this preference
have been updated to return `Source Code Pro` if it's available and
no other preference is set.

Monospace fonts improve the performance of the script editor and code
readability so it's important that the SE defaults to such a font.


livecodeali:
@livecode-vulcan review ok 07dd9da

livecode-vulcan:
💙 review by @livecodeali ok 07dd9dae884d737e8ae088cb5878f22c8fbc862e

PR #1762, opened 2017-09-22
[[ Bug 20391 ]] Fix location of upgrade button

montegoulding:


livecodeali:
I think the button is in slightly the wrong position with this - that's why I did card originally. It looks slightly too high, I think the rect of the toolbar group is a bit odd to get the buttons to appear in the right place

montegoulding:
screen shot 2017-09-22 at 7 08 42 pm
Looks ok to me

livecodeali:
Fair enough

livecodeali:
@livecode-vulcan review ok 6a392c8

livecode-vulcan:
💙 review by @livecodeali ok 6a392c8130681b04cf5de9d42583cd5028d494bb

livecodeali:
screen shot 2017-09-22 at 10 40 36

livecodeali:
Hopefully this is just me

montegoulding:
hmm... little bit odd that we would get different results. Different font metrics?

montegoulding:
One way to solve it would be to set the rect of the graphic to some padded distance from top and bottom of the toolbar group

livecodepanos:
Setting this to WIP until we ensure we handle the cases of the menubar displaying "text only" or "icons only"

montegoulding:
Hmm... good point!

mwieder:
Also on the issue of 'text only' or 'icons only', bug #16880 seems very easy to fix by reducing the stored minHeight of the menubar. This doesn't seem to be set anywhere in the scripts, so my take is that it's in the binary menubar stack, so I can't submit a PR for it. Reducing the minHeight to 23 neatly solves the problem, and has the nice side effect of placement of the Upgrade button in this situation.

livecodeali:
@mwieder The menubar stack is script-only so what you're referring to must be in there somewhere... I'm not sure what you mean by 'the stored minHeight of the menubar though'.

mwieder:
I couldn't find anything that sets the minHeight of the menubar in the script. Looking at the menubar stack in the PI, the minHeight is 32 pixels. If you turn off both text and icons the minHeight is 23 pixels, which hangs the desktop as described in the bug report. I'll dig around in the script some more.

mwieder:
@livecodeali hmmm... I found a place to fix it in the script (maybe the template is just setting the minheight) but the problem of hanging the desktop seems to have disappeared in LC9. Possibly a side-effect of some other change, so I'm going to submit the PR to insure it doesn't reappear.

montegoulding:
@mwieder yes the default minWidth/height from the templateStack is 32 so script only stacks will get that

montegoulding:
screen shot 2017-09-26 at 1 32 55 pm
screen shot 2017-09-26 at 1 32 06 pm
screen shot 2017-09-26 at 1 31 55 pm

montegoulding:
screenshot from 2017-09-26 13-56-36
screenshot from 2017-09-26 13-56-03
screenshot from 2017-09-26 13-55-30
Ubuntu... should be the same on Windows

livecodeali:
@livecode-vulcan review ok 5a9c38d

livecode-vulcan:
💙 review by @livecodeali ok 5a9c38daaed4b5338b362f71d58ca45033089e06

PR #1765, opened 2017-09-24
[[Bug 18305]] Search synonyms without duplicate dictionary entries

bwmilby:
Reworked original fix to no longer require additional entries in the dictionary.
Previous fix required a duplicate dictionary entry for each synonym.

`collectSyntax` - add synonym entries to the string used for searching dictionary
`removeHtmlEntityEscapes` - Turn HTML escape sequences into native characters for search
`addHtmlEntityEscapes` - Turn native characters back into HTML escape sequences after processing
`RegExp.escape` - Escape RegExp special characters for use in searching (added some additional characters)
`dataSearch` - changed to full RegExp special character replacement unless leading space used (same behavior as before, just implemented a little differently), added special case sorting for non-word characters, added line to force list to scroll to the top upon search
`tstate . searched . data . sort` - added synonyms to display name so they would be considered in prioritizing sort (needed since the synonyms will be removed from display name due to another bug fix)
`highlightSearchTerms` - function added to highlight the search term, used on syntax and synonym sections of definition (due to HTML entities, had to convert them before/after and use placeholders for the span tags)
`displayEntry` - added calls to highlight search terms
`$('#ui_filer').keyup` - automatically select the first search item found

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -49,19 +49,55 @@
 	}
 
 
-	// Return all the syntax associated with an entry
+	// Return all the syntax and synonyms associated with an entry
 	// as a (matchable) string
 	function collectSyntax(pEntry)
 	{
 		var tSyntax = pEntry["display name"];
 		$.each(pEntry["display syntax"], function (index, value) 
 		{
-			if (tSyntax != '')
-				tSyntax += ' ';
-				
-			tSyntax += value;
+			tSyntax += ' ' + value;
 		});
+		if(pEntry.hasOwnProperty("synonyms"))
+		{
+			$.each(pEntry["synonyms"], function (index, value) 
+			{
+				tSyntax += ' ' + value;
+			});
+		}
+		tSyntax = removeHtmlEntityEscapes(tSyntax);
 		return tSyntax;
+	}
+	
+	// Replace HTML entity escape sequences with native characters for
+	// searching and highlighting syntax and synonyms.
+	function removeHtmlEntityEscapes(pTerm)
+	{
+		var tTerm = pTerm;
+		tTerm = tTerm.replace(/&amp;/g,"&");
+		tTerm = tTerm.replace(/&lt;/g,"<");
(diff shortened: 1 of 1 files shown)


montegoulding:
@bwmilby great! any chance you could add a screenshot or animated gif of what this looks like?

bwmilby:
screen shot 2017-09-23 at 11 00 00 pm
Here's a shot of a search.  The matched text just has a simple yellow background.

montegoulding:
Nice! And I guess if you type `grc` you get `graphic` with `grc` highlighted in synonyms

bwmilby:
Correct (same for cd, wd, etc.)
I also will select the first entry in the list which makes things like 'wd' more evident since the first entry will have the synonym highlighted.

montegoulding:
Cool well this looks good to me but I'll wait for @livecodeali to take a look before reviewing as it's slightly different to what I discussed with him in terms of the way to let people know that the search term is a synonym. I think what you have done is better than what we discussed though.

bwmilby:
Sounds good.  The highlight code was the trickiest part due to the HTML stuff.  Nice thing about the way I did it makes it simple to change the decoration that is used (could even switch to a div and/or use a css class if we wanted to).

mwieder:
Brian- this looks great. I don't use the dictionary any more, for a number of reasons, but this looks like it will really help anyone who still does.

...and I have too much baggage here to get into the issue of synonyms... :)
(urk... I added a smiley to this and github wouldn't show it)

bwmilby:
newsearch

bwmilby:
Live preview available here:
https://milby.us/lc/docs/api.html

mwieder:
Thanks for the live preview - I hadn't realized you could say 'choose rect tool'.

livecodeali:
I will add my thumbs up here :-) Definitely better than what we came up with!

montegoulding:
@livecode-vulcan review ok bfd5ce5

livecode-vulcan:
💙 review by @montegoulding ok bfd5ce502ffdd5587c40b19eebdd3478b982bf8c

PR #1768, opened 2017-09-25
ensure the menubar height >= the minimum height

mwieder:
Bug #16880 no longer seems to be affecting my system, but this looks like it's just a side-effect of some other change. The problem comes from setting the height of the menubar below the minimum height specified for the stack. Apparently this can be done by script even though the setting prevents it being done manually. My guess is that disallowing it from a script level might have other repercussions, so this patch only affects the menubar stack height and sets the minimum height to 19, which is the height specified for the menubar on OSX with no text and no icons. I'd rather have constants than magic numbers for these values, but I wanted to keep this PR simple.

### DIFF
--- a/Toolset/palettes/menubar/revmenubar.livecodescript
+++ b/Toolset/palettes/menubar/revmenubar.livecodescript
@@ -1,4 +1,15 @@
 ﻿script "revMenuBar"
+// MDW-2017-09-25 [[ bugfix_16880 ]]
+constant kTextOnlyHeightOSX=19
+constant kIconsOnlyHeightOSX=32
+constant kTextAndIconsHeightOSX=47
+constant kTextOnlyHeight=44
+constant kIconsOnlyHeight=59
+constant kTextAndIconsHeight=75
+constant kNoTextNoIconsHeightOSX=0
+constant kNoTextNoIconsHeight=23
+constant kMenubarMinimumHeight=0
+
 local sMenuTarget
 on preOpenStack
    global gREVDontLoadMenus
@@ -6,9 +17,9 @@ on preOpenStack
       put false into gREVDontLoadMenus
       exit preOpenStack
    end if
-   
+
    subscribeMessages
-   
+
    setDefaultText
    setMenuProperties
    generateMenu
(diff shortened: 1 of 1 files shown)


montegoulding:
@mwieder I noticed yesterday that at least cinnamon was not letting me set the width/height of the stack below the minimums via script. So there's probably a general issue here. I agree all these values should be constants.

Could you amend this patch to remove all the whitespace diffs please? It is _very_ hard to find the actual script change in this PR.
```
git checkout -p HEAD^
```
y for all the whitespace patches
```
git add -p
```
again y for all the whitespace diffs
```
git commit --amend
```
At this point if you could add the detail to your commit msg that you have in your PR comment that would be helpful
```
git push -f
```
Also this should be rebased onto develop-8.1 and a release note for the bug added if it's still causing issues on 8.x so once the whitespace diffs are removed if you could:
```
git rebase --onto=develop-8.1 develop
git push -f
```
Then edit the PR to change the target branch.

mwieder:
Wow. will do. I did this quickly and didn't notice the whitespace changes. Which is weird, because all I did was open the file in atom and add two lines and save it. And my git core.autocrlf=false. The 8.1 thing I'll have to check - been a while since I've looked at the 8 series.

montegoulding:
Atom strips trailing whitespace by default. You can add the following to your atom config.cson:
```
  whitespace:
    removeTrailingWhitespace: false
```
Either under `*` or just under `.livecodescript.source`

mwieder:
Hmmm... I'm not averse to removing trailing whitespace. The actual changes are annotated in the usual way with MDW-2017-09-25 [[ bugfix_16880 ]]. I can go back and painstakingly fix the whitespace issue, but I think it's cleaner this way now and in the future.

montegoulding:
It's not painstaking. Just use git commands like I detailed above. It's also not cleaner as when editing the file in the IDE it will add lots of this whitespace back (not all as there appears to be some in odd places) and also will cause unnecessary merge conflicts on this file. Unfortunately if we accept it like this then from then on git blame on most of this file will say `ensure the menubar height >= the minimum height` when the changes hand noting to do with that.

mwieder:
Um... it's a text file. Doesn't have to be edited in the IDE. Point taken about git blame.

PR #1769, opened 2017-09-26
Fix start center field

livecodepanos:
In LC 8.2 DP-1, the script of `revStartCenterBehavior` was refactored, and the `firstRunVisible` command is no longer present. However, the script of `field "firstRunSkip" of group "firstRun" of card "interface" of stack "revStartCenter.livecode"` still refers to it:

```
on mouseUp
   firstRunVisible "false"
end mouseUp
```

Rather than modifying this script (which lives in a binary .livecode file), this patch has re-introduced this command (`firstRunVisible`) in `revStartCenterBehavior.livecodescript`, and modified it accordingly.

### DIFF
--- a/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
+++ b/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
@@ -76,6 +76,11 @@ command initialiseInterface
    end if
 end initialiseInterface
 
+on firstRunVisible pVisible
+   revIDESetPreference "StartCenterClassic", not pVisible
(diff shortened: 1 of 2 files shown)


livecodeali:
Oops! I definitely moved that handler into the script at some point... must have lost it

livecodeali:
@livecode-vulcan review ok b1455de

livecode-vulcan:
💙 review by @livecodeali ok b1455dee68a0fdee93bd485948dfdab68daa3b43

PR #1770, opened 2017-09-26
removed unnecessary trailing whitespace

mwieder:
Now that the menubar has been scriptified and can be edited in a text editor, the trailing whitespace is no longer necessary. This branch does nothing else but remove that. Action: open the file in atom, save.


livecodeali:
@mwieder We don't actually want the trailing whitespace removed, because it gets re-added when the script is edited in the IDE (which is still the most common place to edit it). You might want to tweak the Atom settings for .livecodescript files to not remove the trailing whitespace.

PR #1771, opened 2017-09-27
[[ Bug 20366 ]] Add edition icon and upgrade button to docs

montegoulding:
This patch uses the edition element to add an icon to each non-community
API in the search results list. The same icon use used in the edition
element. If the API is not available in the current edition an upgrade
button is shown.

### DIFF
--- a/Documentation/html_viewer/js/dictionary_functions.js
+++ b/Documentation/html_viewer/js/dictionary_functions.js
@@ -1,4 +1,4 @@
-	var tState = {selected:"",history:{list:[],selected:-1},searched:{},filters:{},filtered_data:[],data:"",selected_api_id:"", sort_type:""};
+	var tState = {selected:"",history:{list:[],selected:-1},searched:{},filters:{},filtered_data:[],data:"",selected_api_id:"", sort_type:"", edition:""};
 	
 	if($.session.get("selected_api_id")) tState.selected = $.session.get("selected_api_id");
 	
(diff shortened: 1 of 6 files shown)


Documentation/html_viewer/js/dictionary_functions.js
```diff
@@ -483,7 +486,29 @@
...
+		else if (tState.edition == "commercial" && tEntryObject["edition"].toLowerCase() != "business")
+			tUpgrade = false;
+		else if (tState.edition == "communityplus" && (tEntryObject["edition"].toLowerCase() != "business" || tEntryObject["edition"].toLowerCase() != "indy"))
+			tUpgrade = false;
+		else if (tEntryObject["edition"].toLowerCase() == "community")
+			tUpgrade = false;
+		
+		if (tUpgrade)
+		{
+			tHTML += ' <a href="#" class="btn btn-default" onclick="javascript:liveCode.showUpgradeOptions(\''
```
livecodeali: Could you make `upgrade` a class and `edition` and `display_name` attributes of the element, and then add an 
```
 		$("body").on("click", ".upgrade", function() {
 			var tEdition = $(this).attr("edition");
                        ...
 		});
```
to the `setActions` function please? It makes it much easier to see what is happening, modify and add other places that need to show upgrade options if we want them

livecodeali: Plus you can gate the call to `showUpgradeOptions` with `isRunningInLiveCodeBrowser()` so there's no error in an external browser

montegoulding: @livecodeali sorry have had my head down for a few days and just read this. Yes to the first comment. Yes to the second but it makes me wonder if we could build the URL locally so the button can work when in an external browser.

montegoulding: OK so I'm just redirecting them to livecode.com for now so it doesn't do nothing if not running in the browser widget.

livecodeali: That's a good idea about making the button work in an external browser yeah...  could just redirect here: https://livecode.com/products/livecode-platform/pricing/

livecodeali:
@livecode-vulcan review ok 17434f2

livecode-vulcan:
💙 review by @livecodeali ok 17434f293789f2c89757b2907c65685ab06ad3d1

PR #1772, opened 2017-09-27
[[ Bug 20475 ]] Improve `Go to definition` for behaviors in use

montegoulding:
This patch changes `seMatchingDefinitions` to use `revBehaviorUses`
to get the first user of a behavior and use that as the available
handler context. This requires some jiggery pokery with the descriptions
array so that the private handlers of the behavior script are included
and those of the user are not.

This patch also replaces the use of `revAvailableHandlers` with the
more efficient `revScriptDescription`.


livecodeali:
@livecode-vulcan review ok 57ac11f

livecode-vulcan:
💙 review by @livecodeali ok 57ac11f2168cf342f8c82ef39d6fff947f23a822

PR #1773, opened 2017-09-27
[[ ScriptExtensions ]] Support script library extension loading

livecodeali:
This patch adds support for loading script library extensions in
the same way as lcb extensions are loaded. Metadata is retrieved
from the docs of the extension.

This patch removes the hard-coding of loading and docs generation
of current script library extensions in the IDE; such hard-coding
will no longer be necessary after companion PRs are merged.


montegoulding:
@livecodeali would it be a good idea to move these to components at the same time? At least in the commercial repo

livecodeali:
I'm not sure... are they components? Are components and extensions the same thing?

montegoulding:
pretty much... at least in the commercial repo anything that we might want to pick and choose which edition it should go in should probably be a component. Whether we need to organise things in such a way in the public repo is debatable though. Ideally we would end up with just one way to define a modular thing though.

livecodeali:
Depends on https://github.com/livecode/livecode/pull/5993

notes/feature-script_extensions.md
```diff
@@ -0,0 +1,88 @@
+# Script Extensions
+Support has been added for installing script libraries contained
+in livecode extension packages (.lce files).
+
+## revLoadLibrary and revUnloadLibrary messages
+
+`revLoadLibrary` and `revUnloadLibrary` handlers should be implemented
```
montegoulding: @livecodeali I wonder if this is an opportune time to rename rev* stuff?

livecodeali: Perhaps yeah... just `loadLibrary`?

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -568,37 +575,46 @@ on revIDEDeveloperExtensionUninstall pPath
...
-   launch document tFile
+   launch document pFile
    if the result is not empty then
       # AL-2015-04-01: [[ Bug 15130 ]] Check to see if the text editor preference is set
       local tEditor
       put revIDEGetPreference("LCB_textEditor") into tEditor
       if tEditor is not empty then
-         launch tFile with tEditor
+         launch pFile with tEditor
       end if
```
montegoulding, line 603: @livecodeali this seems a little odd that the OS pref would take precedence over the local IDE pref

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -568,37 +575,46 @@ on revIDEDeveloperExtensionUninstall pPath
...
    end if
    
    if the result is "no association" or the result is "request failed" then
       answer file "Select text editor..."
       if it is not empty then
-         launch tFile with it
+         launch pFile with it
```
montegoulding, line 609: Shouldn't this set the pref above?

Toolset/libraries/revideextensionlibrary.livecodescript
```diff
@@ -838,35 +862,82 @@ function revIDEExtensionsOrderByDependency pExtensions
...
+   local tExtFile, tType
+   filter tFiles with "*.lcb" into tExtFile
+   put "lcb" into tType
+   if tExtFile is empty then
+      put "lcs" into tType
+      filter tFiles with regex pattern ".*\.livecode(script)?$" into tExtFile
+   end if
+
+   put tExtFile into rSource
+   put tType into rType
```
montegoulding: @livecodeali if `tExtFile` is still empty here this would set `rSource` & `rType` incorrectly. Perhaps a type of `none`?

Toolset/libraries/revideextensionlibrary.livecodescript
```diff
@@ -838,35 +862,82 @@ function revIDEExtensionsOrderByDependency pExtensions
...
+   end if
+   
+   put pFolder & slash & tExtFile into tExtSource
+   if tType is "lcb" then
+      put __fetchModuleData(tExtSource) into tArray
+   else
+      put __fetchScriptLibraryData(tExtSource) into tArray
+   end if
+   
+   put tExtFile into tArray["source_file"]
```
montegoulding, line 864: I think `revIDEExtensionFetchSourceFromFolder` may feasibly set this to a list of files. Perhaps it should be an error if there is something other than a single file found in `revIDEExtensionFetchSourceFromFolder`?

notes/feature-script_extensions.md
montegoulding: hmm.... well script extensions might feasibly be something other than libraries I guess. `loadExtension, unloadExtension` would be nice but then it might be expected to appear in `loadedExtensions`?

All having the same prefix might be nice. Perhaps:
- `extensionInitialize`
- `extensionTeardown`
- `extensionStartupScript`

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
livecodeali: hmm, yeah, might be worth filing a bug report.

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
livecodeali: It does, below, if the launch was successful

Toolset/libraries/revideextensionlibrary.livecodescript
livecodeali: It does do that, above: https://github.com/livecode/livecode-ide/pull/1773/files/fca791decc95eb8c276ae94edbcf14b28baccc4c#diff-5a54288a32499468dd2cab9f178bed1dR891

montegoulding:
@livecode-vulcan review ok 7249bfe

Once merged we will need updates to mergJSON library and tsNet driver to use the new `extension*` API.

livecode-vulcan:
💙 review by @montegoulding ok 7249bfe86813535b0d8b2faab1938d6fa6a84ba0

livecodeali:
@montegoulding We don't currently fork those for develop do we? Probably the best bet is to keep the `rev` ones and add identical `extension` ones until we are no longer building 8.x

montegoulding:
Ah, yes... hmm would it be worth doing something like this for the time being?:
```
try
    send "extensionInitialize" to stack tStackName
catch tError
    send "revLoadLibary" to stack tStackName
end try
```

livecodeali:
Hmm, yeah that might be the best idea

montegoulding:
@livecode-vulcan review ok 4c5349e

livecode-vulcan:
💙 review by @montegoulding ok 4c5349ec415280335bcd6c1898e60e2f5a871262

montegoulding:
@livecode-vulcan review ok 9360372

livecode-vulcan:
💙 review by @montegoulding ok 9360372d72857b93b0b839a9e42d094363ca8df7

montegoulding:
@livecode-vulcan review ok 6d6bedd

livecode-vulcan:
💙 review by @montegoulding ok 6d6bedda8f80222f59e42016c20a071e62f006e7

PR #1775, opened 2017-09-29
fix formatting on two entries

jameshale:
correct entries for dgHilitedLines & dgHilitedIndexes


jameshale:
[Dictionary] update to datagrid.lcdoc

jameshale:
more to do

PR #1785, opened 2017-10-09
Bugfix 20535 & 20536

montegoulding:
I made this a single PR because it's the same bug. I just noticed the suspend one when fixing the suppress on.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -8614,30 +8614,34 @@ on revIDEToggle pProperty
       case "Suppress Messages"
          global gRevSuppressMessages
          put "suppressMessages" into tToggleMessage
+         
+         local tLine
          if not gREVSuppressMessages then
             put true into gREVSuppressMessages
             # Send the suppress messages toggle message before actually suppressing messages
             ideMessageSend "ideToggleChanged:" & tToggleMessage
             set the lockCursor to false
-            repeat for each line l in the pendingMessages
-               if char 1 to 3 of item 2 of l is not "rev" then cancel (item 1 of l)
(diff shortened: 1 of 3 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -9585,36 +9589,36 @@ command revIDESuspendDevelopmentTools
...
          end if
       end repeat
    else
       put the frontScripts into tScriptsList
-      repeat for each line l in tScriptsList
-         if the short name of l is among the items of "revFrontScriptLibrary,revDebugger,revShortCutsLibrary" then
-            put l & cr after tRemoveFrontList
-            remove script of l from front
+      repeat for each line tLine in tScriptsList
+         if the short name of tLine is among the items of "revFrontScriptLibrary,revDebugger,revShortCutsLibrary" then
```
livecodeali: I don't think there's a stack called `revDebugger` anymore, so that can probably be removed!

mwieder: The frontscript is called revdebuggerlibrary now, so if the intent is to remove it from the environment as well the name should be changed here. Also, for consistency in case it matters on some platform, the stack names should be all lowercase.

montegoulding:
I have changed it to `revDebuggerLibrary` and opened a report for a general moomin on this feature to make sure it still does what it's meant to do.

http://quality.livecode.com/show_bug.cgi?id=20623

livecodepanos:
@livecode-vulcan review ok 3b6a36f

livecode-vulcan:
💙 review by @livecodepanos ok 3b6a36f29aa4f77974d5985cbbb74bd946fbf707

PR #1788, opened 2017-10-10
[[ Bug 20217 ]] Indicate layerMode issues when setting

montegoulding:
This patch adds a layerMode setter command so that the effective
layermode can be checked after setting to ensure the user is
informed if it differs from the setting they applied.


montegoulding:
@runrevmark @livecodeali I have left this as WIP mainly because for some reason getting the `effective layerMode` directly after setting the `layerMode` returns the previous value. There must be some refresh point or something. I tried adding a wait and that didn't work. The general idea is to let users know exactly what's blocking their setting.

PR #1791, opened 2017-10-10
[[ Bug 20431 ]] Show ask/answer dialogs in correct location on Mac

montegoulding:
This patch fixes an issue on Mac where if the screenRect has a non-0
top then the dialogs won't be shown at the correct location. The patch
is the same to both stacks:

```
 switch the platform
       case "MacOS"
          local tMacOSX
          put true into tMacOSX
          set the loc of this stack to tStackLoc
-         set the top of this stack to round(item 4 of tScreenRect/5)
+         set the top of this stack to item 2 of tScreenRect + round(tScreenHeight/5)
```


livecodeian:
@livecode-vulcan review ok 3f7b0e3

livecode-vulcan:
💙 review by @livecodeian ok 3f7b0e31bef7204052fcbdaaa1214f3d4bf52582

PR #1794, opened 2017-10-12
[[ Dictionary ]] Tweak filter list layout

livecodeali:
This patch forces the two columns in the filter list to be 50% of
the container width. It also changes the associations list to be
one per line instead of two, since they can be quite long and
it looks bad with huge gaps on one side.

### DIFF
--- a/Documentation/html_viewer/css/lcdoc.css
+++ b/Documentation/html_viewer/css/lcdoc.css
@@ -264,10 +264,16 @@ div.business:before {
 	overflow-y: scroll;
 }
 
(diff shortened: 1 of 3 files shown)


livecodeali:
screen shot 2017-10-12 at 10 09 30

livecodeian:
@livecode-vulcan review ok 8079fd5

livecode-vulcan:
💙 review by @livecodeian ok 8079fd531a43f61e126c1ee78e2b9c994b8eb42c

PR #1799, opened 2017-10-12
[[ ExtensionDocs ]] Don't manually update docs on extension install

livecodeali:
This patch adds a subscription to ideExtensionsChanged to the ide
library, which when received causes the built dictionary data to
be regenerated.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -31,6 +31,7 @@ on revUnloadLibrary
    if the target is me then
       remove the script of me from back
    end if
+   revIDEFinaliseIDELibrary
 end revUnloadLibrary
 
 #############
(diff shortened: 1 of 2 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1513,8 +1518,14 @@ on revIDEInitialiseIDELibrary
    
    ## Initialise property library
    revIDEPropertyLibraryInitialise
+   
+   ideSubscribe "ideExtensionsChanged"
 end revIDEInitialiseIDELibrary
 
+on ideExtensionsChanged
+   revIDERegenerateBuiltDictionaryData
```
montegoulding, line 1526: @livecodeali would it be a good idea to add in a delay and cancel here so if multiple extensions are loaded the docs are regenerated only once?

livecodeali: I think it's maybe not much of an issue, as the time to reassemble the main `.js` is pretty tiny, especially compared to the time taken to generate the json from an `api.lcdoc`

montegoulding: OK, cool

montegoulding:
@livecode-vulcan review ok 322f8ab

livecode-vulcan:
💙 review by @montegoulding ok 322f8ab9de8477e0d6021fb725764d9794e92523

PR #1807, opened 2017-10-20
[[ Bug 20583 ]] Fix dictionary search

livecodeali:
`startsWith` was only introduced in ECMAScript 6, and thus does
not have full browser support. Replace it with `charAt(0) ==`


runrevmark:
@livecode-vulcan review okay 0665409

livecode-vulcan:
💙 review by @runrevmark ok 066540939d4e71af32b5095a373625b9203717e6

PR #1809, opened 2017-10-21
[[ Bug 18199 ]] Add filtering options to standalone builder

livecodeali:
wq7wwamzzf


livecodeali:
I'm not quite sure about the behavior of the platform filters here - should clicking a platform initially _unselect_ it? I think the expected behavior is probably what I've implemented, but then you're in an odd situation where if they are all selected manually, the clicking one again does a different thing (in that situation it really does unselect).

macMikey:
Oh, cool, something to take off my to-do wishlist, now I can just throw ideas at you :-)

Suggestion on platform behavior - check the standalone settings to see which platforms we are going to build for, since inclusions for other platforms aren't necessary.  That's what I was going to do when I was originally thinking about attacking this.  Maybe instead of showing the platforms and having the checkbox, you could have a button that implies "only display platforms that I've selected to build for"

Other things I was thinking about adding:  1) Having column labels, and using those to do sorting of the column (e.g. platform, name, inclusion type, selected/deselected, 2) tooltips for each that help figure out what's in the inclusion.  For example, there is both mergJSON and JSON Library.  One has to read the dictionary to see what does what, and now that we are discussing removing the sections breakdowns, it's may take more work to figure out what's what.  I mistakenly selected the wrong one for an app I was building, this past week, and it took a couple hours to figure out what the heck was wrong.

montegoulding:
😞 binary conflict

macMikey:
So now what?

benr-cogapp:
What macMikey said! This looks like a very useful enhancement - what can be done to get it off the step?

bwmilby:
#1988 will go along well with this addition - will be able to filter platform specific widgets/libraries.

montegoulding:
@bwmilby yes indeed! Nice... perhaps even automagically filter based on the chosen build targets?

Regarding the binary conflict someone on the team is going to need someone to re-implement the UI changes.

PR #1810, opened 2017-10-24
Ensure revUnloadLibrary is sent to the rulers library

livecodepanos:
`tStackName` was `stack revRulersScriptLibrary`, so `send "revUnloadLibrary" to stack tStackName` did not send `revUnloadLibrary` to this stack

### DIFF
--- a/Toolset/libraries/revinitialisationlibrary.livecodescript
+++ b/Toolset/libraries/revinitialisationlibrary.livecodescript
@@ -56,13 +56,11 @@ command revInternal__UnloadLibrary pLibraryName
       return "Library not loaded"
    end if
    
-   local tStackName
-   put sLoadedLibraries[pLibraryName] into tStackName
-   if there is not a stack tStackName then return "library not loaded"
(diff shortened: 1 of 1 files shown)


livecodeali:
@livecodepanos I'm not sure this is correct, I think the idea is that the loaded libraries array keeps a mapping from (potentially informal) library names to stack names - so this change will likely break something else somewhere.

In fact, the following things should work fine:
```
put the name of stack "myStack" into tStackName
answer there is a stack tStackName
send "something" to stack tStackName
```
So I think the problem might be the fact that `tStackName` is `stack revRulersScriptLibrary` instead of `stack "revRulersScriptLibrary"` - that would mean that somehow the array element has been constructed manually rather than using `the name of stack...` although I don't know how that's possible

livecodepanos:
@livecodeali Yes you are right. I have reverted this patch but now I cannot reproduce the bug. Moreover, my previous comment was not correct:

`tStackName` is actually `stack "revRulersScriptLibrary"`

So I am closing this PR and I'll try to reproduce the bug again.

PR #1812, opened 2017-10-25
[[Bug 19969]] PI/Geometry Manager 'Left object' alignment issue

bwmilby:
The PI for GM currently sets the ObjectSide for alignment to the same name as the side which aligns to the opposite side of the target.  If the target object scales, then the intended relation is not preserved.  The solution is to align the object to the nearest/opposite side.

If Btn A is on the left side of Btn B, then when you align the left side of Btn B to Btn A, it currently is calculated against the left side of Btn A.  When looking at it visually, it makes more sense to calculate against the right side of Btn A since that is the edge nearest to Btn B.

Fix adds a function to return the opposite side and changes the property assignment to use the new function.  This function will also be useful when enabling the proportional scaling of objects against the left side of the card (which the library code supports).

### DIFF
--- a/Toolset/palettes/inspector/editors/com.livecode.pi.geometry.behavior.livecodescript
+++ b/Toolset/palettes/inspector/editors/com.livecode.pi.geometry.behavior.livecodescript
@@ -334,7 +334,7 @@ on mouseDown pButton
             default 
                break
          end switch
-         setValue tValue
+         setValue tValue, tKey
       end if  
    end if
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok 26db4d8

livecode-vulcan:
💙 review by @livecodeali ok 26db4d824d7006cfb2d1c00b110cf22323f510fa

PR #1813, opened 2017-10-26
Ensure a handler can be chosen from the SE's "continue" button more than once

livecodepanos:
The `actionEditDebugEntry` command shows a dialog that allows you to choose a handler (and specify params) when pressing the "continue" button of the script editor.

This worked as expected the very first time, but then the dialog was not shown, because `"debugEntryHandler"` was already set (and thus `actionEditDebugEntry` was not called). This was not correct, since one should be able to change the `"debugEntryHandler"` as often as they like.

This patch ensures that `actionEditDebugEntry` is called in any case.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
@@ -2080,25 +2080,16 @@ command actionDebugRun
       revSEGetCurrentObject
       put the result into tObject
       
-      local tHandler
-      put revMetadataGet(tObject, "general", "debugEntryHandler") into tHandler
-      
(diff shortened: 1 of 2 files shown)


montegoulding:
@livecodepanos to me it looks like the original intent was to have the continue button start the current debug entry handler and to set it if it wasn't already set. You can set it via the menu if it's already set to the wrong thing. So this may not be a bug if unless we want to change the behavior. If you want to repeatedly test the same handler with the same inputs then the current behavior would save you entering them again. An alternative might be to pre-populate the dialog with the current values.

livecodepanos:
@montegoulding you are right, this is not a bug. I was not aware of the menu option "Debug -> Entry Point". So I am closing this PR

PR #1815, opened 2017-10-29
[[Bug 20619]] PI/Geometry: Limit Object not selected correctly

bwmilby:
The PI/Geometry tab does not correctly select the Limit Object check box when viewing a control.
The code sets the check box value based on field values that are populated after the check box is set.
Moving the code corrects the issue.


montegoulding:
@livecode-vulcan review ok 5f730a6

livecode-vulcan:
💙 review by @montegoulding ok 5f730a63d38c3060d077fbd1377b200eac712ce2

PR #1816, opened 2017-10-31
[Bug 20621] fix some indentions for uncommon If forms

BerndN:
two cases
1
      if a then 
         if b then get c else get d -- OFFENDING LINE
         else -- should be on the same indention level as first if
            doSomethingElse 
         end if

2
      if a
      then put b else put c -- OFFENDING LINE
  
   if a -- should be on the same indention level as first if
   then put b


BerndN:
I will never learn how to do the bugfix note correctly :(

montegoulding:
@BerndN could you add your cases to `tests/scripteditor/_indentation_tests/if.livecodescript` then we do a test PR against the engine repo using your patch and ensure that none of the indentation tests are broken.

BerndN:
@montegoulding  is that correct?

montegoulding:
Looks like it. Thanks @BerndN 

The test PR is here https://github.com/livecode/livecode/pull/6087

montegoulding:
@livecode-vulcan review ok abe664a

The test PR failed on travis but the failures were unrelated to this patch and the current and added indentation tests passed.

livecode-vulcan:
💙 review by @montegoulding ok abe664acb2d551b29198ecad310dbd382d794b2f

PR #1817, opened 2017-11-02
[[Bug 17287]] Image filename property is not relativized with the PI

bwmilby:
There is an option in preferences to "always use absolute file paths for images" but the PI does not change how it operates based on that setting.  This change modifies the file behavior to convert paths to relative if that setting is not made to use absolute.  It will impact both images and other media (anything in the PI that uses the file behavior).

### DIFF
--- a/Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
+++ b/Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
@@ -56,6 +56,13 @@ on mouseUp pButton
       put the result into tResult
       
       if there is a file tResult then
+         if the cImageFileRelative of stack "revPreferences" is not false then
+            local tStackPath, tFilePath
+            set the itemDelimiter to "/"
+            put item 1 to -2 of the filename of stack revTargetStack(line 1 of (the selObj)) into tStackPath
+            put revCalculateRelativePath(tStackPath, tResult) into tFilePath
+            if there is a file (tStackPath & slash & tFilePath) then put tFilePath into tResult
+         end if
          put tResult into field 1 of me
          valueChanged
       end if
--- a/Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
+++ b/Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
@@ -56,12 +56,15 @@ on mouseUp pButton
       put the result into tResult
       
       if there is a file tResult then
-         if the cImageFileRelative of stack "revPreferences" is not false then
+         if revIDEGetPreference("cImageFileRelative") is not false then
             local tStackPath, tFilePath
             set the itemDelimiter to "/"
-            put item 1 to -2 of the filename of stack revTargetStack(line 1 of (the selObj)) into tStackPath
-            put revCalculateRelativePath(tStackPath, tResult) into tFilePath
-            if there is a file (tStackPath & slash & tFilePath) then put tFilePath into tResult
+            put item 1 to -2 of the filename of stack revTargetStack(line 1 of the cSelectedObjects of this stack) into tStackPath
+            if tStackPath is not empty then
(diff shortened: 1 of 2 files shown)


Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
```diff
@@ -56,6 +56,13 @@ on mouseUp pButton
       put the result into tResult
       
       if there is a file tResult then
+         if the cImageFileRelative of stack "revPreferences" is not false then
```
montegoulding: This should use `revIDEGetPreference("cImageFileRelative")`

Toolset/palettes/inspector/editors/com.livecode.pi.file.behavior.livecodescript
```diff
@@ -56,6 +56,13 @@ on mouseUp pButton
...
+            local tStackPath, tFilePath
+            set the itemDelimiter to "/"
+            put item 1 to -2 of the filename of stack revTargetStack(line 1 of (the selObj)) into tStackPath
```
montegoulding: There is a problem here in that the PI may not be operating on one of the selected objects. Indeed there may be no selected objects in which case there will be an error thrown on this line. Try `the cSelectedObjects of this stack`

montegoulding:
@bwmilby this seems reasonable to me. The main thing I'm unsure about is using an image related pref here where it may not be setting an image path. There's the possibility to move this relativisation to a preference setter command so it just applies to image filename... @livecodeali what do you think?

livecodeali:
Yeah I think we would need to do that for backwards-compatibility... One thing we may want to do is add a "Relativize file path" checkbox to the `com.livecode.pi.file` template group, then allow this to be done generally

bwmilby:
I did a global search for "cImageFileRelative" and the only hits I found were in the preferences and my addition.  (I also used a text editor to search that has found things inside binary stacks before.) 
 Is it possible that it is no longer used?  If so, could it be renamed to just cFileRelative with an adjustment to the label in the pref panel?

One of the suggestions that was made in the discussion was a global property preference (like the one I used) but also an option in the PI.  That would use the template mod that was suggested and the default value would reflect the preference.

montegoulding:
@bwmilby I think we should leave the pref name as is and add other prefs for other file path properties if necessary. I also think that rather than implementing this in `com.livecode.pi.file` we should add a setter to image filename in `com.livecode.interface.classsic.Image.tsv` say `filename::revIDESetImageFilename`

Then add something like this to the IDE library
```
command revIDESetImageFilename pObjectID, pProperty, pFilename
   if revIDEGetPreference("cImageFileRelative") is not false then
       local tStackPath, tFilePath
       set the itemDelimiter to "/"
       put item 1 to -2 of the filename of stack revTargetStack(pObjectID) into tStackPath
       if tStackPath is not empty then
           put revCalculateRelativePath(tStackPath, pFilename) into tFilePath
           if tFilePath begins with "./" then delete char 1 to 2 of tFilePath
           if tFilePath ends with tResult then
               --full path contained in relative path, likely a different volume
                --in any case, need to return absolute path
           else if there is a file (tStackPath & slash & tFilePath) then
                put tFilePath into pFilename
           end if
        end if
    end if
    set the filename of pObjectID to pFilename
end revIDESetImageFilename
```
Also I think this should be submitted against develop-8.1 or maybe 8.2

bwmilby:
@montegoulding If we define a setter like this, would that mean that the image files would always get changed to relative or does that just effect the PI?  (Looks like it just is for the IDE/PI reading the .md file in the .tsv folder)
Should I try to rebase this or just close and submit a new PR?  I could cross reference this one for the discussion.  (Since the .tsv change is currently just on 8.1, I would do it there).

montegoulding:
@bwmilby well I would suggest it would be appropriate for `revIDEImportControl` to re-use the handler when importing a referenced image but it wouldn't do so until we patched it.

bwmilby:
1848 merged.  Closing this since it is now an obsolete duplicate.

PR #1819, opened 2017-11-03
[[ Tests ]] Wrap indentation tests in handlers

montegoulding:
This patch wraps the indentation tests in handlers with a comment after
the structure to test that the indentation after the structure is correct.

@BerndN noticed the tests wouldn't pick up one of the issues he recently fixed. This patch should make sure it would be picked up in the future. The issue was the test:

```
if foo()
then code else code
```
Did not pick up the previous behavior of:
```
on foo
   if foo()
   then code else code
-- comment
end foo
```

### DIFF
--- a/tests/scripteditor/_indentation_tests/if.livecodescript
+++ b/tests/scripteditor/_indentation_tests/if.livecodescript
@@ -1,36 +1,57 @@
-if foo() then
-   -- code
-else if bar() then
-   -- code
-else
-   -- code
-end if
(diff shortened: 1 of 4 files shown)


livecodeali:
@livecode-vulcan review ok 337069f

livecode-vulcan:
💙 review by @livecodeali ok 337069fd1fd7176c6b27c29c79ef9385c66579ba

PR #1822, opened 2017-11-04
[[Bug 13825/18097]] Profile Manager/Geometry Manager on Mobile

bwmilby:
Profile Manager (PM) and Geometry Manager (GM) do not work on mobile due to missing library code in the standalone that is built.

GM depends on a handler present in the common library (REVTargetStack).  That library was not being loaded on mobile.  There is a good bit of code in that library that is not applicable to mobile.  ~~This PR includes a new library of the common code that is pertinent to the mobile platform (revmobilelibrary).~~

PM depends on getprop/setprop handlers in the common library.  ~~They are included in this new library.~~

The standalone settings excluded the profile manager option when iOS was selected.  That exclusion has been removed.

~~In this first commit, every handler name from the common library was left as a comment in the mobile library.  I used # for the ones that I'm confident could just be removed.  I've left the others with -- for discussion.  The intent would be to remove the comment lines before final merge.~~

Last 2 commits remove added file and make some updates to existing common library.


Toolset/libraries/revmobilelibrary.livecodescript
```diff
@@ -0,0 +1,621 @@
...
+#--function macVersionLessThan pLeft, pRight
+
+# Parameters
+#   pObject : either a long id or a valid reference to an object.
+# Returns
+#   The "rugged" id of the object. This can be used the same as a long id,
+#   but is slightly more resistant to stuff like objects being grouped, renamed etc.
+#   The script editor uses rugged ids to store object ids when there is a chance that the object
+#   could be renamed or grouped etc. 
+function revRuggedId pObject
```
bwmilby: Reading comments in the code, it appears as if this function is not really needed except for inside the IDE (there is even at least one case where a check for development environment is done and this is only called if so).  If that is indeed true, then this function could be stripped down for the mobile library and just return the long id of the object.

montegoulding: @bwmilby the remote debugger needs this

montegoulding:
@bwmilby I'm not convinced this is a good move given the code duplication. I would prefer to see the common library patched with either mobile implementations of common library handlers that are not currently supported on mobile or just something like:

```
if the environment is "mobile" then
   throw "not implemented"
end if
```
At some point this library will need a thorough cleanup (perhaps a split up too) and move to an extension with embedded docs etc so I'm concerned we may be adding to the work that needs to be done to do that if we add a mobile version. There's lots I don't like about this library. The use of shell and applescript irks me along with stuff like the `mouseDoubleUp` handler that I really have no idea what that was implemented for as there's no comment.

Toolset/libraries/revmobilelibrary.livecodescript
bwmilby: This code is not currently available on mobile unless remote debugger pushes the common library somehow.

montegoulding: yes it adds it in as a dependency along with the debugger library and the metadata library

bwmilby:
@montegoulding I think it would be better to split the library than to include all of the desktop only code in a mobile application (partly from a code size perspective).  I think the new library would be a good starting point for the common library and everything else could be in a desktop library.  There may be some additional handlers that would make sense to move over (the way it sits now makes it easy to see all of the handlers in the original library).  I did perform some cleanup on the code I moved over but the only thing mobile specific was elimination of checks for development (and use of rugged id).  Changing that back would be easy.  This could also be the first step in getting to an extension by splitting out the stuff that truly is "common" between most platforms.

Much of that old file code that used applescript was probably needed for legacy Mac OS.  I would imagine that OS X would not need any of that type of code, but don't have any experience to know for sure.

Due to the way the code is split between IDE and Engine, both PRs will need to be updated if the library name that needs to be included on mobile changes.

Toolset/libraries/revmobilelibrary.livecodescript
bwmilby: Looking at that code (debugger library), i can see the call, but is the whole function really needed on mobile?  There are places in code where if not in development environment then the long ID is used.  Reading the comments looks like the code is needed to account for objects in groups while in edit group mode.  My guess is that at some point it was added to the common library due to multiple dependencies (profile library?).  Could it be simplified in the common library to check for development and pass and if not just return long ID?
```
# simplified function for standalones where edit group mode can't cause a problem
function revRuggedId pObject
   if the environment is "development" then
      pass revRuggedID -- let revBackscript handle the call
   else
      return the long id of pObject
   end if
end revRuggedId
```

montegoulding: It is used considerably more by the remote debugger.

Toolset/libraries/revcommonlibrary.livecodescript
```diff
@@ -137,32 +140,30 @@ on revSetWindowsShellCommand
...
-   # because its located in revBackscript (which is not included in standalones). As we don't require a
-   # rugged id in standalones anyway, just revert to the old behavior in standalone environment.
+getProp revProfile
+  return the cREVGeneral["profile"] of the target
+end revProfile
+
+getProp cREVGeneral[pWhichProp]
    local tTarget
    if the environment is "development" then
+      # OK-2008-04-14/OK-2008-06-02 : The long id will return an invalid reference
```
montegoulding: @bwmilby this is a bit odd... just using `the target` below instead of `tTarget` should work in both cases as `the target` is a direct object reference. Indeed in both cases here `the target` is clearly a valid object because otherwise you could not get the long id of it. It should also be marginally faster as we would not be getting the long id pr revRuggedId string and then parsing it again to get an object ref.

montegoulding: Note that `the target` can be both a function to get the name of the object and a chunk destination type to directly reference the object (like `me`). Here we are using it as a chunk destination type in all cases.

Toolset/libraries/revcommonlibrary.livecodescript
```diff
@@ -1352,6 +1354,14 @@ function revFontGetSystemFont
          return "Helvetica"
          break
          
+      case "iphone"
```
montegoulding, line 1300: Hmm... I wonder if this should just return `(Default)` on all platforms now.

Toolset/libraries/revcommonlibrary.livecodescript
```diff
@@ -1418,6 +1436,12 @@ private function _revEnvironmentPlatform
       case "Win32"
          put "Windows" into tPlatform
          break
+      case "iphone"
```
montegoulding, line 1382: we should add in html5 while we are at it.

Toolset/libraries/revcommonlibrary.livecodescript
montegoulding: perhaps we need a way to query the font name and size for the themed fonts

Toolset/libraries/revcommonlibrary.livecodescript
bwmilby: @montegoulding I did notice that the setProp handler did not use the rugged id.
In the getProp handler, originally it used "the target" before the rugged id was obtained.  I moved the rugged id code to the beginning of the handler and optimized the revUniqueID code.
To be sure I understand, your suggestion is that `the target` be used instead of `tTarget` throughout the getProp handler.  That would be an easy change.
I wonder why this was inserted back in 2008 though.  I don't know enough about rugged ids to identify a case where the current code is necessary.

Toolset/libraries/revcommonlibrary.livecodescript
bwmilby: Is that something that would need to be added to the engine?

Toolset/libraries/revcommonlibrary.livecodescript
bwmilby: will do (updated in the next commit)

Toolset/libraries/revcommonlibrary.livecodescript
montegoulding: @bwmilby It could just be that at one point someone decided that getting the long id and putting it into tTarget was a good idea (not sure why) and later someone fixed the case that it didn't work in edit group mode.

Toolset/libraries/revcommonlibrary.livecodescript
montegoulding: yes it is

bwmilby: @montegoulding I just did a search for "revFontGetSystemFont" and the only hits are in the common library.  I also used a tool that has produced hits within binary stacks before that also came up empty.  I also searched for "revFontApplySystemStyle" (only place the above function is used) and it is only used in 2 files:  revaboutcardbehavior.livecodescript (1) and revonlinelibrary.rev (4).  From that, it seems that this is more of an IDE thing and not needed/used in stand alone apps.

Toolset/libraries/revcommonlibrary.livecodescript
```diff
@@ -224,6 +214,7 @@ setProp cREVGeneral[pWhichProp] pWhichProfile
       unlock messages
       exit cREVGeneral
    end if
+   
    if pWhichProp is "breakPoints" then
       
       pass cREVGeneral
```
bwmilby: Since we are doing cleanup, should we remove this whole section of code (`breakPoints`)?  The pass could be handled at the top with the initial check for the properties that are handled. 230/221 to 271/262 could be removed as the code is never reached at a minimum.

Toolset/libraries/revcommonlibrary.livecodescript
montegoulding: Hmm... it could be an undocumented handler that is still used by some users... I'm not sure

Toolset/libraries/revcommonlibrary.livecodescript
montegoulding: Yes you are right this can all be removed and just remove breakpoints from the condition at the top. This should have been cleaned up when the pass was put in.

montegoulding:
@livecode-vulcan review ok cf8c72b

livecode-vulcan:
💙 review by @montegoulding ok cf8c72b73133eb4e9a7185187d711dbe2aa2a87e

PR #1823, opened 2017-11-06
[[ TSV ]] Use .tsv extension for tab delimited files

montegoulding:
It appears the files need some schmoozing that I didn't bank on so I'll come back to this later


montegoulding:
Replaced by https://github.com/livecode/livecode-ide/pull/1826

PR #1824, opened 2017-11-07
[[Bug 20645]] Create mobile scroller each time DataGrid is opened

trevordevore:


trevordevore:
Note to reviewer: I haven't tested this yet as I only have XCode 9.1 installed on my computers and LC doesn't support Xcode 9.1 yet. Bug report can be found here:

http://quality.livecode.com/show_bug.cgi?id=20645

Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
```diff
@@ -210,12 +195,34 @@ private command _Initialize
       unlock messages
       
       put true into sInit
+   else
```
montegoulding: @trevordevore thanks for this. It looks good the only issue is I think this needs to be guarded against execution on non-mobile platforms. So `else if the environment is "mobile" then`

trevordevore: Fixed.

montegoulding:
@livecode-vulcan review ok a6ddd9a

Thanks @trevordevore

livecode-vulcan:
💙 review by @montegoulding ok a6ddd9ac230e8dcb4abd81d9de39efaa0da88bc5

PR #1825, opened 2017-11-07
[Bugfix 20514] Extensions: Add a way to launch sample stacks

runrevelanor:
Sample stacks can be launched from the Extension Manager. Launched stacks
have their filename set to empty to the user is propted to choose a
location to save to is they save they stack. This ensures the original
sample stack is always opened from the Extension Manager and the user can
make changes and save the stacks.

It is assumed that the sample stacks will be in a "samples" folder in the
extension folder.


montegoulding:
@runrevelanor just incase you missed it there's a conflict on this. Unfortunately GitHub doesn't notify about conflicts

livecodeali:
@runrevelanor You will also need to modify the `revIDEDeveloperExtensionLibrary` script so that the samples folder gets included in the extension package (just look for everywhere that the resources folder is dealt with and do similar with the samples folder)

montegoulding:
@livecodeali Just checking you have seen this PR has conflicts

livecodeali:
@montegoulding Ta, fixed now.

montegoulding:
@livecode-vulcan review ok 84644fc

livecode-vulcan:
💙 review by @montegoulding ok 84644fc5b9226eb944a51777b9c561d3a944f55d

PR #1827, opened 2017-11-10
[[ Editions ]] Add script_profiler property

montegoulding:
This patch just means that when we BFS a commercial edition the IDE doesn't get confused if we aren't running business


montegoulding:
needs rebase

runrevmark:
@livecode-vulcan review ok 49d38e9

livecode-vulcan:
💙 review by @runrevmark ok 49d38e910a863c405c4132a4b64a0279764b3fd9

PR #1829, opened 2017-11-11
[[Bug 20666]] Property Inspector / Geometry: Remove All button

bwmilby:
The "Remove All" button on the Geometry tab of the Property Inspector currently has no effect.

The first change adds a parameter to the setValue call (where the button click is handled) to remove all geometries and expands that handler to match (depends on an earlier PR that is merged already).

The second change is to add a command to clear geometry keys that are not needed.  If no geometry sides are set, then the other associated settings are removed.  This new command is called from the revIDESetGeometry handler.

The third change is to add checks to prevent the revCalculateGeometryDistances from calculating distances for sides that are not needed.


montegoulding:
@livecode-vulcan review ok 12c321c

livecode-vulcan:
💙 review by @montegoulding ok 12c321c72d721c4264558f1eb3031a2ff2ecfb2f

PR #1838, opened 2017-11-21
[[ ExtensionStandaloneSettings ]] Add mechanism for extensions standalone settings

livecodeali:
This patch adds several features to enabled extensions to specify standalone settings in their metadata. These appear as cogs in the inclusions list, which then pop up a property inspector-style sheet of properties that can be set.
screen shot 2017-11-21 at 15 20 25


montegoulding:
Hmm... We really need to make search for inclusions a button you click every now and then to update this list

livecodeali:
How does that relate to this? (I agree, just wondering if you've noticed an aspect of it that this exacerbates)

montegoulding:
Just that you won't see the cog unless you are not searching for inclusions

livecodeali:
ah, yes, good point

montegoulding:
That probably doesn't need to cause this to be WIP though. We just need a small change I think. Make search for inclusions boolean with checkbox and make it add to existing inclusions rather than clobber them then make it save inclusions after the search so they appear in the list next time. Perhaps adding a search now button would also be good.

montegoulding:
@livecode-vulcan review ok ae3c945

livecode-vulcan:
💙 review by @montegoulding ok ae3c9454e67d44292d76e9949cc63fb6721c24b2

PR #1842, opened 2017-11-25
[[Bug 20684]] Inconsistent arrow key behavior in the dictionary

bwmilby:
Up/down keys were not able to navigate between dictionary entries (always went to $ when they worked).  This was due to the way the `entry_previous` and `entry_next` functions were attempting to get the currently displayed entry ID.  This fix uses the history list to get the current entry ID.

These 2 functions were separated by a bit of code before, but are now consolidated to be together for easier maintenance in the future.

Previously, the up/down keys always navigated the filtered list.  Adjusted the code to check for a search term and use that list if available.

The history functions needed bounds checking to prevent the `tState.history.selected_index` from getting set to invalid values.

### DIFF
--- a/Documentation/html_viewer/api.html.template
+++ b/Documentation/html_viewer/api.html.template
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok 5bcff2e

livecode-vulcan:
💙 review by @livecodeali ok 5bcff2eee70e4635e2e525a69f916a5e97afe5de

PR #1848, opened 2017-12-01
[[Bug 17287]] Relativize image filename property in the PI

bwmilby:
There is an option in preferences to "always use absolute file paths for images" but the PI does not change how it operates based on that setting.  This change adds a setter for image filename that will relativize the path unless `cImageFileRelative` is false (default value is unset but shown as true in preferences, so check for it not being false vice checking for true).

Original PR for reference/prior discussion:
https://github.com/livecode/livecode-ide/pull/1817


livecodeali:
@livecode-vulcan review ok d4232ab

livecode-vulcan:
💙 review by @livecodeali ok d4232abcdf8b97aa5b82b1d68fbfb9c62714512b

PR #1853, opened 2017-12-11
Line Ending Conversion Correction

bwmilby:
Native Mac line endings are not being converted (CRLF and LFCR are converted to LF but CR was not touched).
Changed to use the LF constant to make it more readable.
Also changed numToChar to numToCodepoint since the former function is deprecated.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -5012,9 +5012,10 @@ end arrayRecurse
 private function __escapeStringAndConvertLineEndings pString
    replace "\" with "\\" in pString
    replace quote with ("\" & quote) in pString
-   replace (numToChar(13) & CR) with (CR & numToChar(13)) in pString
-   replace (CR & numToChar(13)) with CR in pString
-   replace CR with "\n" in pString
+   replace (numToCodepoint(13) & LF) with (LF & numToCodepoint(13)) in pString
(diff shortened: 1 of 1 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -10679,10 +10679,10 @@ private function escape pString, pConvertLineEndings
    replace quote with ("\" & quote) in pString
    
    if pConvertLineEndings is true then
-      replace (numToChar(13) & CR) with (CR & numToChar(13)) in pString
-      replace (CR & numToChar(13)) with CR in pString
+      replace CRLF with "\n" in pString
+      replace numToCodepoint(13) with "\n" in pString
```
montegoulding, line 10683: @bwmilby would need to examine all the code paths leading here but I'm wondering if `pConvertLineEndings` is false here we should be escaping `numToCodepoint(13)` with `\r`

bwmilby: @montegoulding Function only used once (line 11050) and it has "true" as the convert value.  It probably makes sense to put an else clause in though.

montegoulding:
@livecode-vulcan review ok e91b84f

livecode-vulcan:
💙 review by @montegoulding ok e91b84f4d57060250b2d0c78b0d3014ee6e6fd8f

PR #1855, opened 2017-12-14
[bug 17819] enable cmd+c in dictionary

BerndN:
cmd+c did not work when anything else in IDE was selected, text, objects etc.
http://quality.livecode.com/show_bug.cgi?id=17819

### DIFF
--- a/Toolset/libraries/revshortcutslibrary.livecodescript
+++ b/Toolset/libraries/revshortcutslibrary.livecodescript
@@ -156,6 +156,13 @@ on commandKeyDown pWhich
                send "mouseUp" to btn "clear" of stack "revMessageWatcher"
             end if
             break
+         case "C"
+            -- bug 17819 commandkey + c in dictionary not working
+            if revTargetStack(the focusedObject) is "revDictionary" then
+               send "revMenubarEditMenuPick" && quote &"Copy Text" & quote to \
+                     stack "revIDEMessageHandlerLibrary"
+            end if
+            break
          default
             pass commandKeyDown
       end switch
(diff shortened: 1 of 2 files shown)


montegoulding:
@BerndN I'm thinking it's probably better to handle `commandKeyDown` in `revDictionaryBehavior`

BerndN:
@montegoulding I tried that and it works much to my surprise by simply putting 

on commandKeyDown pWhich
end commandKeyDown

into revDictionaryBehavior

I am a bit at a loss why this is so. Without understanding why I would not want to submit this.

Of course 

on commandKeyDown pWhich
   if pWhich is "C" then
      send "revMenubarEditMenuPick" && quote &"Copy Text" & quote to \
            stack "revIDEMessageHandlerLibrary"
   end if
  pass commandKeyDown
end commandKeyDown

works also.

montegoulding:
@BerndN it seems probable to me that the browser is handling the shortcut and the IDE is clobbering it by handling `commandKeyDown`. It could be that such a thing isn't cross platform though. If this works I'd run with that:

```
on commandKeyDown pWhich
  if pWhich is "C" then
     copy
  else
     pass commandKeyDown
  end if
end commandKeyDown
```

BerndN:
@montegoulding 

```
on commandKeyDown pWhich
  if pWhich is "C" then
     copy
  else
     pass commandKeyDown
  end if
end commandKeyDown
```

works

What do you want me to do? Close this pull-request and do a new one?

montegoulding:
whatever is easier for you. If you are using the web interface then it's probably easier to close and reopen I guess.

BerndN:
OK, then I will close this one and do a new one
Thanks

PR #1856, opened 2017-12-14
[bug 17819] enable cmd+c in dictionary

BerndN:
cmd+c did not work when anything else in IDE was selected, text, objects etc.
http://quality.livecode.com/show_bug.cgi?id=17819


montegoulding:
Ah the only issue now is the other PR was against develop-8.1 which is probably where we want this

BerndN:
@montegoulding 
Should I close this and do it against develop-8.1 ?

montegoulding:
yes unfortunately I don't think there's a way to rebase via the web interface

PR #1865, opened 2017-12-18
[[ Bug 19946 ]] Add active extension folders preference

livecodeali:
Any folder containing a folder containing an extension selected
for loading by the user is added to the cDeveloperExtensionsFolders
preference. This patch adds a cDeveloperExtensionsActiveFolders
preference which controls which of the above folders are actually
searched for extensions in the extension builder. This preference
can be set using the cog at the top-right of the extension builder
stack.


runrevmark:
@livecode-vulcan review ok 9dc037f

livecode-vulcan:
💙 review by @runrevmark ok 9dc037f163a3b9fb18a7c0617b2d7c752ffb7858

PR #1869, opened 2017-12-18
Fix initial setting of LCB editor pref

bwmilby:
The `revIDEDeveloperExtensionEditScript` handler will open an LCB file in the default editor.  It uses `the result` to catch failures to open with a default editor and set a pref.  The problem is that after evaluating the next `put` (getting the editor pref), the original value is lost.  This means that if the pref has not already been set, then the file won't get opened.

This change implements a new variable `tResult` to carry forward the result though the function.

### DIFF
--- a/Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
+++ b/Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
@@ -593,7 +593,7 @@ on revIDEDeveloperExtensionEditScript pFolder
       end if
    end if
    
-   if tResult is "no association" or the result is "request failed" then
+   if tResult is "no association" or tResult is "request failed" then
       answer file "Select text editor..."
       if it is not empty then
          launch tFile with it
--- a/Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
+++ b/Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
@@ -593,19 +593,21 @@ on revIDEDeveloperExtensionEditScript pFolder
       end if
    end if
    
-   if tResult is "no association" or tResult is "request failed" then
+   if tResult is "no association" or tResult is "request failed" \
+         or tResult is "no such program" then
       answer file "Select text editor..."
-      if it is not empty then
-         launch tFile with it
+      put it into tEditor
+      if tEditor is not empty then
+         launch tFile with tEditor
          put the result into tResult
          if tResult is empty then
             # AL-2015-04-01: [[ Bug 15130 ]] If the new text editor launch was successful, set the preference
-            revIDESetPreference "LCB_textEditor", it
+            revIDESetPreference "LCB_textEditor", tEditor
          end if
       end if
    end if
    
-   if tResult is not empty then
+   if tResult is not empty and tResult is not tEditor then
       __revIDEDeveloperExtensionSendError "Could not open" && tFile & ":" && tResult
    end if
 end revIDEDeveloperExtensionEditScript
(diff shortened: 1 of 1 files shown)


Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -577,28 +577,31 @@ on revIDEDeveloperExtensionUninstall pPath
...
       local tEditor
       put revIDEGetPreference("LCB_textEditor") into tEditor
       if tEditor is not empty then
          launch tFile with tEditor
+         put the result into tResult
       end if
    end if
    
-   if the result is "no association" or the result is "request failed" then
+   if tResult is "no association" or the result is "request failed" then
```
livecodeali: There's a second `the result` here that should be replaced

livecodeali:
@bwmilby thanks for this!

bwmilby:
@livecodeali Glad to help.  Not sure how much help this will need merging into 9 since this code was factored out to a new function call.

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -577,28 +577,31 @@ on revIDEDeveloperExtensionUninstall pPath
...
       answer file "Select text editor..."
       if it is not empty then
          launch tFile with it
+         put the result into tResult
       end if
    end if
    
-   if the result is not empty then
+   if tResult is not empty then
       __revIDEDeveloperExtensionSendError "Could not open" && tFile & ":" && the result
```
livecodeali: oops, missed one here too!

livecodeali:
Actually, do we think that the LCB_textEditor preference should take precedence over file association? I think @montegoulding has suggested this in the past.

Toolset/libraries/revidedeveloperextensionlibrary.livecodescript
```diff
@@ -577,29 +577,32 @@ on revIDEDeveloperExtensionUninstall pPath
...
       end if
    end if
    
-   if the result is not empty then
-      __revIDEDeveloperExtensionSendError "Could not open" && tFile & ":" && the result
+   if tResult is not empty then
+      __revIDEDeveloperExtensionSendError "Could not open" && tFile & ":" && tResult
    else if it is not empty then
       # AL-2015-04-01: [[ Bug 15130 ]] If the new text editor launch was successful, set the preference
       revIDESetPreference "LCB_textEditor", it
```
montegoulding: I would like to see this moved up to after line 600 with:
```
if tResult is empty then
        # AL-2015-04-01: [[ Bug 15130 ]] If the new text editor launch was successful, set the preference
        revIDESetPreference "LCB_textEditor", it
end if
```
I don't feel safe that `it` will be empty if there were already an association.

bwmilby:
If the preference were to take precedence over the file association, then I think there needs to be a way to set/reset the pref in the IDE somewhere (probably a good idea anyway).  Could be a menu option on the cog of the Extension Builder stack (palletActions widget) since this is the only place it is used.

livecodeali:
Yeah, that's a good idea. I'm not sure exactly what the best way of populating the list of possible programs would be though.

bwmilby:
Doing some testing with the additions and found an issue.  The result is being set to the app name.

montegoulding:
@livecode-vulcan review ok 9a63f32

livecode-vulcan:
💙 review by @montegoulding ok 9a63f3201796b7993ff3d4b215859e9afc013921

PR #1872, opened 2017-12-19
[[ CEF ]] Use CEF by default on linux

montegoulding:
The updated version of CEF now works on a significant number of
distributions. This patch will allow CEF to be used by default
unless a user explicitly sets `$LIVECODE_USE_CEF` to `0`.


runrevmark:
Reviewed by me.

PR #1873, opened 2017-12-20
remove any generated doc files after running the IDE

mwieder:
After compiling and launching the compiled build, the data directory is generated inside the resources directory. Git is not happy about this, so this patch updates the .gitignore file to ignore the generated files and the directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,8 +66,8 @@ environment_log.txt
 
 # Documentation #
 #################
-Documentation/html_viewer/resources/data/api/*.js
-Documentation/html_viewer/resources/data/guide/*.js
+# Documentation/html_viewer/resources/data/api/*.js
+# Documentation/html_viewer/resources/data/guide/*.js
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -68,6 +68,7 @@ environment_log.txt
 #################
 Documentation/html_viewer/resources/data/api/*.js
 Documentation/html_viewer/resources/data/guide/*.js
+Documentation/html_viewer/resources/data/
```
montegoulding: @mwieder how about changing the above lines to a catch all `Documentation/html_viewer/resources/data/*/*.js`

mwieder:
The latest LC9dp build added more subdirectories, so even just ignoring all .js files won't do it. Best to ignore the whole data subdirectory if you don't want to scrub it.

montegoulding:
Closed as https://github.com/livecode/livecode-ide/pull/1931 did the same thing

PR #1875, opened 2017-12-20
[[ DataGrid 2]] Assorted bug fixes.

livecodemichael:
Manually send reorder move. This effectively throttles the mouse move
message (which was interfering with animations in certain situations)
and also means that we allow widgets to be used as reorder controls
(since widgets don't pass mouse move by default).

Make sure to set the horozontal position of the reorder control in the
LayoutControl handler. This ensures the reorder control is in the
correct position after scrolls and resizes.

Extend the reorder scroll check to beyond the rect of the data grid.
This means users can drag above or below the group and the scroll will
continue.

Make sure we send the animation pulse message at the correct intervals.

Use the correct animation properties for animating the reoder control.

Correctly check the bounds on reordering, preventing reodering line
numbers not in the datagrid.


livecodepanos:
@livecode-vulcan review ok 3cb73a5

livecode-vulcan:
💙 review by @livecodepanos ok 3cb73a503cb106766954f2a4d954c79fa9e83597

PR #1877, opened 2017-12-21
[[ DataGrid 2 ]] Fix drags to cooperate with the mobile scroller.

livecodemichael:
Previously, a data grid's mobile scroller was getting in the way of
reorder and swipe actions.

To fix for reordering, we disable the scroller when the user clicks
on a reorder control (and re-enable it when the reorder completes).

Swipes are a little trickier. We track dragging on mouse down, but
disabling the scroller at this point will prevent all scrolling.
Instead, we attempt to determine if the user is trying to drag the
row before disabling the scroller.


livecodepanos:
Tested locally and it fixes the problem.

@livecode-vulcan review ok ac1beee

livecode-vulcan:
💙 review by @livecodepanos ok ac1beee1dd113e203b6715fe472cf2fa88656bd5

PR #1880, opened 2018-01-08
[[ Bug 20830 ]] Add iconGravity to PI

montegoulding:


Toolset/resources/supporting_files/property_definitions/propertyInfo.tsv
```diff
@@ -86,6 +86,7 @@ hotspots	Hotspots	Quicktime	com.livecode.pi.string	true	true
 hoverIcon	Hover icon	Icons	com.livecode.pi.imageid	true	false		0	
 hScrollbar	Horizontal scrollbar	Basic	com.livecode.pi.boolean	true	false		false		
 icon	Icon	Icons	com.livecode.pi.imageid	true	false		0	
+iconGravity	Gravity	Icons	com.livecode.pi.enum	true	false			,left,top,right,bottom,topLeft,topRight,bottomLeft,bottomRight,center,resize
```
montegoulding, line 89: @livecodeali it would be nice to allow the word `empty` here or `(empty)` or something. I'm not sure how deep or risky that change might be.

livecodeali: Hmm, yeah... I'm not sure either!

livecodeali:
What's the logic in making the readable name "Gravity Icons"? Isn't it just "Icon Gravity"?

livecodeali:
PS I've always thought the `iconGravity` editor should be a square split into 9 panels and an icon that you could drag from one panel to another... never worked out really how to do that or indeed what to do if the button doesn't currently have an icon set. Anyway, better to have it in there as a standard enum than not at all!

montegoulding:
There's a tab between `Gravity` and `Icons` so it's named `Gravity` and placed on the `Icons` pane. Oh for yaml ;-)

I like your idea for the future... we could just use a dot if there's no icon.

livecodeali:
doh!

livecodeali:
@livecode-vulcan review ok 7507e53

livecode-vulcan:
💙 review by @livecodeali ok 7507e53f120028001c8f6ecf6de13d2cdec734cf

PR #1883, opened 2018-01-13
bug # 20868 -- enable/disable button "replace" based on search location

MartinKoob:
Replacing of found items in Script Editor is only available when Location option is 'Current Tab' or 'All Tabs' so the button "replace" of stack "revSEFind">Stack "main" should only be enabled then.  For all other Locations the "replace" button should be disabled.   Also when replace is not available and disabled the label of the button should be set to empty.  It can be set back to item 1 when the user changes back to location 'Current Tab' or 'All Tabs'.


livecodeali:
@livecode-vulcan review ok c3adc0c

livecode-vulcan:
💙 review by @livecodeali ok c3adc0ca9fdffe858ff3aed47cf6f420c0cf0f88

PR #1892, opened 2018-01-29
[[ Bug 20851 ]] Fix case/break completion

montegoulding:
This patch fixes the completion of `case` statements with `break` where
an existing `break` statement already exists. The depth was not being
calculated correctly so that if multiple `switch` structures were in the
same script the second and subsequent structures added extra `break`
statements when hitting return after the `case` statement.

Additionally this patch ensures that `case` is only completed with `break`
if there is no other code for the case as it is unhelpful to put `break`
before the code under any circumstances.


livecodeali:
@livecode-vulcan review ok 5149261

livecode-vulcan:
💙 review by @livecodeali ok 5149261777e4c8775805a4c3f2677f6300cccadc

PR #1895, opened 2018-01-29
[[ Bug 20920 ]] Launch SE when SOS dragged onto Mac dock icon

livecodeali:


montegoulding:
This won't work if LC is not already open because the IDE engine automatically opens the first stack in the command args. I vaguely recall I have a couple of stale PRs to remove that behavior from the IDE engine so we can control when/how it opens.

montegoulding:
https://github.com/livecode/livecode-ide/pull/1311
https://github.com/livecode/livecode/pull/4348

livecodeali:
👍  Cool

PR #1901, opened 2018-01-31
[[ Bug 20332 ]] Handle unset app id and set for user

montegoulding:
This patch handles the case where the user has not yet set an app id
and additionally offers to set the app id to a string that matches the
chosen profile.


livecodepanos:
@livecode-vulcan review ok e0dfc2c

livecode-vulcan:
💙 review by @livecodepanos ok e0dfc2c0c3b24b056dd5b034541bac6f9c9f6f74

PR #1903, opened 2018-01-31
[[ Bug 20929 ]] Output to module.<module_version>.lcm when compiling

livecodeali:
In 9.0 DP 11, we attempt to recompile extensions that are not the correct format for the current version. Unfortunately, this overwrote the existing module.lcm file. This patch fixes things so that we start compiling modules into `module.<module_version>.lcm`.


Toolset/libraries/revideextensionlibrary.livecodescript
```diff
@@ -13,6 +13,52 @@ on extensionFinalize
...
+   put item 1 to -2 of tTempFile into tFolder
+   put "module com.livecode.dummy" into tModule
+   put return & "end module" after tModule
+   put tModule into url("file:" & tTempFile)
+   revIDEExtensionCompile tTempFile, "", tFolder, "module.lcm"
+   put __CompiledModuleVersion(tFolder & slash & "module.lcm") \
+         into sModuleVersion
+end __EnsureModuleVersion
+
+private function __CompiledModuleVersion pModule
```
montegoulding, line 33: @livecodeali should we have an engine function to get this do you think?

montegoulding:
@livecode-vulcan review ok 00c5829

livecode-vulcan:
💙 review by @montegoulding ok 00c5829e2c6d7e7f66c77e2b0c3ee3bc3fef5682

Toolset/libraries/revideextensionlibrary.livecodescript
mwieder: Is that necessary when you can just parse the first few bytes of the .lcm file?

livecodeali: There's a slight subtlety here I suppose with the fact that the output format of the lc-compile bundled with the IDE is not necessarily that of the engine, as they are separate programs - one of them might have been replaced. So it *might* make sense to also have that (essentially the VM version) as an engine function and make sure that they match before doing anything. OTOH maybe we shouldn't bother too much about the case that one or other program has been replaced!

montegoulding: I was thinking there might be environments where you have access to the engine and a package containing compiled modules for multiple versions but not have lc-compile installed. Perhaps in a self updating standalone or something?

PR #1912, opened 2018-02-15
[[ Bug ]] Fix extension 'file' property

livecodeali:
Commit 00c5829 erroneously stored the full path to the module file
in the 'file' property - everywhere that uses the property prepends
the source folder before use, so just store the actual filename.


runrevmark:
@livecode-vulcan review ok aee6bb0

livecode-vulcan:
💙 review by @runrevmark ok aee6bb0c0dede940503de5decc347aaff049823d

PR #1915, opened 2018-02-22
[[ CodeLibrary ]] Fix library mapping for installed extensions

montegoulding:
In theory all code libraries should be loaded by the IDE from the engine
folder. In practice we only do that for a few libraries at the moment
so this patch allows the library mapping to check the engine folder first
then fallback to the code folder.


livecodeali:
Any idea why Git thinks this is such a big change? It's a lot easier to review with ?w=1 :-D

montegoulding:
It's just because what was previously in an `else` block is not now so half the handler has shifted left a bit ;-)

livecodeali:
@livecode-vulcan review ok 789bcb0

livecode-vulcan:
💙 review by @livecodeali ok 789bcb06c47961d6170d2727a3b1d9abc4c73cfa

PR #1929, opened 2018-03-09
[[ DataGrid2 ]] Allow swipe and hold

montegoulding:
This patch allows a user to swipe and reveal the action button and
then hold the touch for a period of time. If they release the touch
while the button is fully exposed then it remains revealed. Previously
if they did not release the touch within the swipe detection time the
action button would be hidden.


livecodemichael:
@livecode-vulcan review ok 49f2834

livecode-vulcan:
💙 review by @livecodemichael ok 49f2834b656946bc06dffd5a2b7fcd34d384caa1

PR #1941, opened 2018-03-21
[[ Bug 21095 ]] Load required extensions on startup

montegoulding:
This patch both ensures that the menu option to not load an extension
on startup in the extension manager is removed from IDE dependencies and
also that IDE dependencies will be loaded regardless of the setting if it
happened to be set in another way or a widget becomes a dependency in a
new version.


livecodeali:
@livecode-vulcan review ok 102476a

livecode-vulcan:
💙 review by @livecodeali ok 102476abc63506ed4f0866003f00a0e592cb56c3

PR #1943, opened 2018-03-21
[[ Bug 19504 ]] Allow tab in text property of PI

montegoulding:


livecodeali:
You are abusing the `options` field a bit here, although I'm not sure there's another option beyond having a `textwithtabs` editor or something. Or declaring the editors as `textwithtabs` and adding a hack to the group behavior like with styledtext. Hmmm

montegoulding:
Hehe... well what other kind of options would we want for a text field?

livecodeali:
@livecode-vulcan review ok e39d852

livecode-vulcan:
💙 review by @livecodeali ok e39d852896b389be3aa8ea8bf5a94b4ad2717c93

PR #1948, opened 2018-03-23
[[ Bug 20094 ]] Execute msg box if it compiles

montegoulding:
This patch ensures that if the message box script compiles ok and
we fail to find any improvements by autocompleting the script it is
still executed.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -11512,7 +11512,11 @@ command ideExecuteScript pScript, pObject, pDebugMode, @rValidScript
    end if
    
    if tValidScript is empty then
-      return tOriginalCompileError for error
+      if tOriginalCompileError is not empty then
+         return tOriginalCompileError for error
+      else
+         put pScript into tValidScript
+      end if
    end if
    unlock screen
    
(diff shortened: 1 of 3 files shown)


livecodeali:
Test build, as message box changes can be fragile
https://github.com/livecode/livecode/pull/6465

montegoulding:
ooh.... thanks for that... it appears to have broken half the tests! Will run locally and see what I can figure out

Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -11511,8 +11511,11 @@ command ideExecuteScript pScript, pObject, pDebugMode, @rValidScript
       end if
    end if
    
-   if tValidScript is empty then
+   if tValidScript is empty and \
```
montegoulding: I think this should be:
```
if tValidScript is empty then
     if tOriginalCompileError is not empty then
           return tOriginalCompileError for error
     else
           put pScript into tValidScript
     end if
end if
```

livecodeali:
@livecode-vulcan review ok bd9fc27

livecode-vulcan:
💙 review by @livecodeali ok bd9fc27dcbbac16bbeb17861e628d6c8d1404680

PR #1954, opened 2018-03-29
[[ Bug 21122 ]] Update windowBoundingRect

montegoulding:
This patch ensures that the windowBoundingRect is updated when the screen
configurations change and when the menubar and tools palette are moved. It
also ensures that script editors are resized to fit within the window when
the screens change.


livecodeali:
@livecode-vulcan review ok a95287b

livecode-vulcan:
💙 review by @livecodeali ok a95287b9d4e885249c6a1eadb83f419e52d16017

PR #1956, opened 2018-04-09
[Bug 21174] Ensure "Sample Stacks" stack does show up in menu "Windows"

livecodepanos:


livecodeali:
@livecode-vulcan review ok f86a6fd

livecodeali:
In order to prevent the need for modifying this in two places in the future, I suggest refactoring slightly to have a function 

```
function revIDEStackIsIDEWindow pStackName
   if pStackName begins with revIDEScriptEditorPrefix() or pStackName is "revDictionary" \
      or pStackName is "revResourceCenter" or pStackName is "revOnline" then
      return true
   end if

   return false
end revIDEStackIsIDEWindow
```
which can then be called from both `revIDEStackIsWindow` and `setupWindowMenu`

livecodeali:
@livecode-vulcan review ok 75f7b92

livecode-vulcan:
💙 review by @livecodeali ok 75f7b92c08199ac5d4ee120ddf9b972e23973d4f

PR #1960, opened 2018-04-12
[[ Bug 21172 ]] Fix revMail on mobile

bwmilby:
Adding the common library to mobile caused the revMail function to break.
`pass revMail` was added to the handler, but the engine did not support that method.

Replaced the `pass` with a call to `mobileComposeMail` to restore functionality

On iPhone this should be a good solution (looking at the source code).  The calls for revMail are much simpler for Android, but the end result should be the same.


livecodeali:
@livecode-vulcan review ok 61bb868

livecode-vulcan:
💙 review by @livecodeali ok 61bb86842ab7e128d30fbb578a495cb776c72a1a

PR #1962, opened 2018-04-17
Fully correct header markdown syntax

jameshale:
some headers also were of the form "###Heading" rather than "### Heading"
This corrects this as well as incorporates previous change from Ali.


montegoulding:
@jameshale it looks like this PR was done against develop then submitted against develop-9.0. @livecodeali's PR is against develop-9.0 so I think it would be best to close this, merge Ali's PR then you can make the patch to just the headers that aren't correctly formed on develop-9.0.

jameshale:
ok

PR #1966, opened 2018-04-20
[[ Bug 21202 ]] Fix deselection of next find after replace

montegoulding:
This patch adds a method to replace text in the script editor without
calling selection update or chaging the selection at all. The current
selection update request was causing find results to be cleared between
replace and go to next and the subsequent replace.


livecodeali:
@livecode-vulcan review ok 9421a12

livecode-vulcan:
💙 review by @livecodeali ok 9421a12ee4bf99acc29d061fb40698af2d355701

PR #1970, opened 2018-04-28
Add codeFolding to IDE

BerndN:
This will add optional code folding to the script editor. It adds a field "Folding" to the right of field "Numbers" in which folding status is shown (open, folded). Clicking on a line with a symbol for folding will toggle folding status for that struct.

Code folding is an option appended to the Edit Menu when in Script Editor. Can be turned on and off there in the same session of editing. Folding preference is saved when quitting Livecode.

Keyboard shortcuts are command K for Collapse all handlers, command U for Unfold will unfold all handlers and other folded structs.
Furthermore command-clicking on the first line of a foldable struct will toggle folding state for that struct.

Foldable are all handlers, if, else, repeat, switch, case, try and "#<"
#< is begin of a block comment and has to be closed with #</ on an arbitrary line.

Nested foldables structs can be folded manually and unfolding the parent struct will not unfold the substruct.

Folding slows down the Script Editor, depending on your hardware and the length of the script.

Up to 5000 lines is ok, beyond that it gets a bit sluggish. This affects mainly making a new line since the script has to be rescanned for foldables and changing tabs in Script Editor. Typing is not affected.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
@@ -29,12 +29,19 @@ command resize
    put tMyRect[3] - tMyRect[1] into tMyWidth
    put tMyRect[4] - tMyRect[2] into tMyHeight
 
-   local tGutterWidth
+   local tGutterWidth, tFoldingWidth
+   
+   if sePrefGet("editor,folding") then
+      put sePrefGet("editor,foldingwidth") into tFoldingWidth
+   else
+      put 0 into tFoldingWidth
+   end if
+   
    if sePrefGet("gutter,linenumbers") then
-      put max(formattedWidth of field "Numbers" of tControls["gutter"], \
+      put max(formattedWidth of field "Numbers" of tControls["gutter"] + tFoldingWidth, \
          kMinGutterWidth) into tGutterWidth
    else
-      put kMinGutterWidth into tGutterWidth
+      put kMinGutterWidth + tFoldingWidth into tGutterWidth
    end if
 
    local tFindHeight
@@ -1097,6 +1104,22 @@ on mouseDown pButtonNumber
       pass mouseDown
    end if
    
+   if sePrefGet("Editor,folding") then
+      if the commandKey is down then
+         local tLineText
+         local tAllowedLineStarts
+         put "on,private,function,command,setProp,getProp,after,before,if,repeat,switch,case,try,else" into tAllowedLineStarts
+         put line sLastClickLine of field "Script" of me into tLineText
+         if word 1 of tLineText is among the items of tAllowedLineStarts or (word 1 of tLineText begins with "#<" and char 3 of word 1 of tLineText <> "/") then
+            if token 1 of tLineText is "else" and token -1 of tLineText <> "else" then
+               exit mouseDown
+            end if
+            folding
+            exit mouseDown
+         end if
+      end if
+   end if
+   
    prepareContextMenu
    showContextMenu
    pass mouseDown
@@ -1401,6 +1424,43 @@ private command findText pText, pGlobal, pStartFrom, pEndAt, pSearch
       end try
    end repeat
    delete the last char of tResult
+   
+   if sePrefGet("Editor,folding") then
+      if item 1 of line - 1 of tResult > 0 then
+         local tStartLine, tEndLine, tCounter, tFoldStart, tFoldEnd, tNoOfLines
+         put the number of lines of char 1 to (item 1 of line - 1 of tResult) of pText into tStartLine
+         
+         if the hidden of line tStartLine of field "Script" of me then
+            put the number of lines of pText into tNoOfLines
+            lock screen
+            
+            -- search backwards until not hidden
+            put tStartLine - 1 into tCounter
+            repeat with i = tCounter down to 1
+               if NOT (the hidden of line i of field "Script" of me) then
+                  put i into tFoldStart
+                  exit repeat
+               end if
+            end repeat
+            
+            -- search forward
+            put the number of lines of pText into tNoOfLines
+            put tStartLine into tCounter
+            put tNoOfLines into tFoldEnd
+            repeat with i = tCounter to tNoOfLines
+               if NOT (the hidden of line i of field "Script" of me) then
+                  put i - 1 into tFoldEnd
+                  exit repeat
+               end if
+            end repeat
+            
+            unfoldAll tFoldStart, tFoldEnd + 1
+            synchFoldedFieldNumbers
+            unlock screen
+         end if -- if the hidden of line tStartLine of field "Script" of me
+      end if -- if item 1 of line - 1 of tResult > 0
+   end if -- if sePrefGet("Editor,folding")
+   
    return tResult
 end findText
 
@@ -1433,7 +1493,12 @@ command findMarkResult pStart, pEnd, pInteractive
    # The field is scrolled by the minimum amount required to bring the current search
    # result fully into view.
    local tHeight
-   put the formattedHeight of char 1 to pEnd of field "Script" of me into tHeight
+   
+   if not sePrefGet("Editor,folding") then
+      put the formattedHeight of char 1 to pEnd of field "Script" of me into tHeight
+   else
+      put the formattedTop of char pEnd of field "Script" of me - the top of field "Script" of me + the vScroll of field "Script" of me into tHeight
+   end if
    
    # If tHeight is before the beginning of the visible area then scroll back until its the first line
    local tScroll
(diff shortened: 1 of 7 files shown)


Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -1097,6 +1104,22 @@ on mouseDown pButtonNumber
       pass mouseDown
    end if
    
+   if sePrefGet("Editor,folding") then
+      if the commandKey is down then
+         local tLineText
+         local tAllowedLineStarts
+         put "on,private,function,command,setProp,getProp,after,before,if,repeat,switch,case,try,else" into tAllowedLineStarts
```
mwieder, line 1111: Could make tAllowedLineStarts a constant and save the time of initializing a variable each time.

BerndN: thanks Mark, I guess there is more to come...

mwieder: this is a great PR though. Thanks for doing this.

bwmilby:
What about non-inline block comments?  If `/* ... */` spans multiple lines, that would seem to be an ideal thing to fold.

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -2919,3 +2984,1206 @@ private command addStack pStack, pObeyDontSearch, pMarked, pIgnorePasswordProtec
...
+                  end if
+               end if
+               
+               put tCounter & cr after tCollect -- regular if
+               next repeat
+               
+            end if -- (aLine contains "then") and not (token - 1 of aLine = "then")
+         end if --token 1 of aLine is "if
+         
+         if token 1 of aLine is "else" and not (token -1 of aLine is "else") then next repeat
```
mwieder: This line prevents "else if" clauses from being recognized. Commenting it out gives a triangle to "else if" lines.
(see below at line 3729)

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
```diff
@@ -2919,3 +2984,1206 @@ private command addStack pStack, pObeyDontSearch, pMarked, pIgnorePasswordProtec
...
+            exit repeat
+         end if
+      end repeat
+      
+      return tStartLine, tLineCounter - 1
+      exit parseStructs
+   end if -- handler
+   
+   -- conditional if
+   if token 1 of tLineText is "if" or (token 1 of tLineText is "else" and token - 1 of tLineText is "else") then
```
mwieder: To fold "else if" clauses, remove the "token -1 etc part. Could just replace with
if token 1 of tLineText is among the items of "if,"else" then
(see above at line 3621)

BerndN: This is a deliberate design decision. I am not sure one should fold "else if" because those are usually short and it would fractionate the folding of an "If" struct even more.

But if the consensus is that it should be foldable then of course your change would make that possible.

BerndN: @bwmilby folding of mulitline comments was an early request for folding. I hesitated to implement it because first I wanted to fold actual code. Then the problem for multi-line "/* */" comment is that there are no clear rules, people nest them which is actully not recommended but does compile. Then the positioning can be anywhere not only at the beginning of a line.

Because of that I decided to put that on hold and do that maybe on a next version if there is a considerable demand for it.

bwmilby: @BerndN I’m thinking mostly about the possibly large block comments before each handler (documentation).  I think that it would be beneficial to include these- just stipulate that it only folds this type initially (where the start/end occur at the beginning of a line - first token)

mwieder: I agree about the usefulness of block comments. These can get rather long.
...and nesting block comments isn't (or at least shouldn't be) allowed. Attempts to nest block comments in a handler properly result in a compiler error. Nesting them outside a handler for some reason doesn't cause an error (which may be a bug in the parser), but apparently anything outside a handler is treated as a nonexecutable comment.

runrevmark: Yes @mwieder is correct - block comments are C-style, so don't nest. In terms of 'what goes on outside of a handler' - then the parser tends to just skip to something it can pass as a variable or handler definition IIRC... Which is why you can do things like:
```
local sVariable

-- on thisHandlerWillBeIgnored
...
end thisHandlerWillBeIgnored

on thisHandlerWontBeIgnored
end thisHandlerWontBeIgnored
```
I suggested 'fixing' that once - but there were quite a few people who sent evils via the power of email so it is as it has always been!

bwmilby: Outside of handlers does kind of make sense for nesting block comments.  That way you can use a block comment to remove an entire handler even if it contained block comments.

Having read Marks reply then, it just looks like they nest outside handlers.

mwieder: Regarding folding "else if": what I did in the glx2 editor is
"if" statements fold to the corresponding "end if" for clarity.
"else" and "else if" statements fold to the next "else..." clause or to the "end if" if there are no more.
Not saying this is The Way It Should Be, just that that was my design decision treating "else" and "else if" clauses as equal at the same level, and within the enclosing "if/end if" clause.

runrevmark: @bwmilby They don't nest outside of handlers - it is just that things which don't look like handler/var definitions are ignored - whatever they are. I'm pretty sure the handler list parser checks to see if the first token of a line is in the SP_HANDLER token table, and if not just skips to end of line. So in the case of a nested block-comment:
```
/*A fobar /* something
else */B */
on foo
end foo
```
The block comment will be tokenized from A to B which leaves ` */` on the line, which isn't a handler token (on,local,global etc.) and so is just skipped.

bwmilby: Right.  What I meant is that you can “nest” a block comment around a single handler and it will execute as if it had nested since the parser will ignore the invalid stuff between where the comment actually ended and the start of the next handler.  Probably better to just comment out the definition statement.

Although you could end up with some unintended script local variables.

BerndN: @mwieder, @bwmilby
thanks for the comments and suggestions
I will try to implement the suggested additions.
Not sure how to change the current pull-request to add them.
Will find out.

bwmilby: Just need to do another commit to your branch.  It will automatically show up on the PR.

BerndN: @bwmilby I am working in the web interface for github. As far as I understand changing a currently submitted pull request will result in just showing the differences from the original pull request. Maybe it is easier to delete those parts of the pull-request and add them anew?

bwmilby: It will be this evening, but I will take a look.

livecodeali:
Hi @BerndN, I have taken the liberty of rebasing this on develop to remove the conflicts - I hope that's ok. If you want to make tweaks to this PR, you should be able to just do them in the normal way via the web interface, as long as you are on the patch-3 branch (i.e. navigate to files in `https://github.com/BerndN/livecode-ide/tree/patch-3/Toolset/`). When you make a commit there, it will be added to this PR. 

I think
> changing a currently submitted pull request will result in just showing the differences from the 
> original pull request

is not true... although admittedly I have not done much in the web interface myself.

Anyway, the reason I was looking at this is that we are preparing various things for the first DP of 9.5, and I was wondering if you wanted to try and get it merged :-) It would be a very cool feature to add!

BerndN:
@livecodeali I will try to incorporate my recent changes to Folding into this pull-request.
If I can't manage that I will do a new pull-request against develop and close this one.

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: fixed it

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: added slash asterisk folding

Toolset/palettes/script editor/behaviors/revseeditorbehavior.livecodescript
BerndN: made it a constant

BerndN:
@livecodeali  I updated the pull-request to the latest features of code folding. (It took me several attempts for some unfortunately)
Then I copied the seven affected .livecodescript files and installed them into LC 9.0.4 RC 2 as a round-trip test.
It worked. Code folding was working as expected. I also turned on grevdevelopment and no errors reported. 

So I think I will not do a new pull-request but recommend this one for review.

BerndN:
had to add empty quoted literals to foldAll.
Otherwise strange things in quoted literals (especially /* or */) could cause malfunctions

mwieder:
9.5 seems to be a dead end. Can this be merged into the 9.6 stream?

PR #1971, opened 2018-04-29
[[ Bug 21245 ]] SE indent errors with inline block comments

bwmilby:
There are 2 related errors in the `lineStripComments` /` __StripComment` handlers in the `revsecommoneditorbehavior.livecodescript` code.

The first error is that "/*" is assumed to run to the end of the current line.  Since it is possible to have multiple inline block comments on a single line, a recursive call to `__StripBlockComment` is used to remove these comments.

The second error is that the loop that goes through the current line does not exit when a true comment to the end of the line is encountered.  The loop uses quotes to ensure that it isn't capturing comment characters in a string, but fails to exit.  Added an additional return value (pass by reference variable) indicating comment status of the current chunk to better facilitate knowing when to exit the loop.

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsecommoneditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsecommoneditorbehavior.livecodescript
@@ -334,7 +334,7 @@ end __GetPreference
 
 -- Text Formatting Code
 command textFormatInitialize
-   -- permit use in environment witn no dependencies
+   -- permit use in environment with no dependencies
    local tTabDepth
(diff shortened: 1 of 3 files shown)


montegoulding:
@bwmilby could you add some indentation tests for the fix to `tests/scripteditor/_indentation_tests`

montegoulding:
@livecode-vulcan review ok 6373c36

livecode-vulcan:
💙 review by @montegoulding ok 6373c36f8d69c48737d04a1c5a036867f25086da

PR #1974, opened 2018-05-14
[[ PERF ]] Ide performance issues

montegoulding:


Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
```diff
@@ -667,8 +667,8 @@ private command __UpdateIntelligenceDisplay
    else
       local tControl
       put the name of it into tControl
-      set the label of button "Open Stacks" of group "Stacks" of me to tIntelligence & ": " & tControl
-      enable button "Open Stacks" of group "Stacks" of me
+      set the label of button "Open Stacks" of group "Stacks" of card "Single Line" of me to tIntelligence & ": " & tControl
+      enable button "Open Stacks" of group "Stacks" of card "Single Line" of me
```
livecodeali, line 671: I think this should be either Single Line or Multiple Lines depending on which is the current card

livecodeali: Perhaps we should have a behavior above the single and multiple line card behaviors that deals with this instead, but that's beyond the scope of this PR ;-)

Toolset/palettes/message box/behaviors/revmessageboxsinglelinemessagebehavior.livecodescript
```diff
@@ -22,8 +22,13 @@ on enterInField
...
 end enterInField
 
+local sColorizeMsg
+
 on textChanged
-   dispatch "revSEColorizeField" to stack "revSEUtilities" with the long id of me
+   cancel sColorizeMsg
+   get the long id of me
+   send "revSEColorizeField it" to stack "revSEUtilities" in 100 milliseconds
+   put the result into sColorizeMsg
```
livecodeali, line 31: I wonder if it would be a good idea to make this a public handler in revSEUtilities and for it to store an array of colorize messages parameterized by the long id. Then we could also have the `100 milliseconds` bit as a constant in that script and people would be able to use it to colorize their fields without having to set up a send/cancel scenario

livecodeali: (By no means essential to this patch ofc)

montegoulding: I was actually considering making the single line msg box use the common editor behavior rather than do this but it would have involved significant enough changes that it would be better to do in develop.

Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
montegoulding: The group is shared but it is only on some cards. We just need to make sure we reference one of the cards its on here and it will work fine. The issue is only if the user is on some other card and the message comes in.

livecodeali:
@livecode-vulcan review ok 8ffaf7b

livecode-vulcan:
💙 review by @livecodeali ok 8ffaf7b1c43d9689dd31d8ff884d7fcdd9f02fab

PR #1978, opened 2018-05-27
[[ Bug 20672 ]] Update property inspector once for multiple objects

montegoulding:
Previously `idePropertyChanged` would be sent for each of the selected objects
causing the property inspector to be updated for each. This patch will ensure
it is only updated once for multiple objects.


livecodeali:
@livecode-vulcan review ok 2d73ed8

livecode-vulcan:
💙 review by @livecodeali ok 2d73ed8182287a3aae7278d36e4ed656e0d57c1f

PR #1981, opened 2018-06-03
[[ Bug 21232 ]] Fix wandering breakpoints

montegoulding:
This patch fixes wandering breakpoints when undoing a paste or rapidly
adding or removing lines via return or delete. It ensures that we
immediately update breakpoints unless the line count as not been changed
by the edit.


livecodepanos:
Tested locally and it fixes the original problem, as well as various other use-cases. Thanks @montegoulding 

@livecode-vulcan review ok 7c94eda

livecode-vulcan:
💙 review by @livecodepanos ok 7c94eda4884ccf87af9f98f092b7b411e2d8cec5

PR #1987, opened 2018-07-24
[[ Bug 21427 ]] Allow nesting of Data Grid Behaviors

bwmilby:
If a developer needed to insert handlers to override any of the
default DG behavior, they were unable to do so using a behavior.
Updated the initialization code to be able to handle one or more nested
behaviors between the DG group and the DG library.

Script local `sResourceStack` created to retain the location of the library.
The value is initialized in the `_Initialize` handler.  All references to
`_ResourceStack()` were normalized and changed to the new variable (4 cases
used `stack _ResourceStack()` and the rest omitted `stack` which is included
in the returned value).  Cases where a local was declared for the value were
changed to use the script local.

The `_ResourceStack()` handler looks at the behavior of `me` first and if it
does not find an expected object (`button "Data Grid" of group "Behaviors"`)
then it will try to next level behavior until there are no more to try.


Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
```diff
@@ -7475,6 +7469,16 @@ end PrintKeys
 private function _ResourceStack
    local theStack, theCharNo
    put the behavior of me into theStack
+   
+   repeat
+      if there is a button "Data Grid" of group "Behaviors" of \
```
trevordevore: @bwmilby This needs a couple of corrections. 

1. `trueword -1 of theStack` won't work for stack names with spaces. `trueword -1 of "My Stack"` would be `Stack`. 
2. You need to check if `theStack is empty` before testing a property of the stack. Otherwise you would end up with an error because you are referencing a property for a non-existent object.

I also came across a situation today that requires a further change. I'm working on a sample Levure project for SQL Yoga. I wanted a custom script only stack behavior for a Data Grid group. The way to specify the Data Grid behavior in the script only stack is as follows:

```
script "ProjectsPeople Behavior" with behavior "RevDataGridLibraryBehaviorsDataGridButtonBehavior"
```

While the behavior is valid the `_ResourceStack` function won't work because "RevDataGridLibraryBehaviorsDataGridButtonBehavior" doesn't have the resources the data grid needs. Something like this should fix all problems though. I hard code the "revDataGridLibrary" reference. Perhaps we could just get away with doing that. 

```
private function _ResourceStack
  if there is a stack "revDataGridLibrary" then
    return the long id of stack "revDataGridLibrary"
  else
   local theStack, theCharNo
   put _StackOf(the behavior of me) into theStack
   
   repeat forever
      if theStack is empty or there is a button "Data Grid" of group "Behaviors" of \
            theStack then
         exit repeat
      else
         put _StackOf(the behavior of theStack) into theStack
      end if
   end repeat
   
   return theStack
  end if
  
end _ResourceStack


/**
* \brief Returns the stack reference of a control.
*
* \param pControl A reference to a control.
*
* \return A reference to the control stack.
*/
private function _StackOf pControl
  local theCharNo

  put the long ID of pControl into pControl ## force the long id

  if word 1 of pControl is not "stack" then
    put offset(" of stack ", pControl) into theCharNo
    if theCharNo > 0 then
      delete char 1 to (theCharNo + 3) of pControl
    else
      put empty into pControl
    end if
  end if
  return pControl
end _StackOf
```

bwmilby: 1.  `stack value(word -1 of theStack)` should work for stacks with spaces in their names.  We can't actually put the resolved stack name into the variable at that point since we are possibly following the behavior chain from object to object.
2.  fixing (as a rev... stack the errors were ignored)
3.  The issue is that you are referencing the script only stack behavior directly instead of the behavior button, but the hard code reference solution should be fine.  That will make it quicker when there are nested behaviors.  If the name ever gets changed, then the fall back code should take care of it.

trevordevore:
👍

PR #1988, opened 2018-08-03
[[Feature]] Add OS and Platforms to LCB Metadata

bwmilby:
Update *Extending LiveCode* guide with note and example about adding
metadata for OS and/or Platforms if the module uses code that is only
available on specific machines.

Updated `__fetchMetadatum` to include the new keys in the manifest.xml
file.


montegoulding:
@livecode-vulcan review ok 79f4d8c

livecode-vulcan:
💙 review by @montegoulding ok 79f4d8caf05c091ba9d6226bed8ba714121e07e6

PR #1991, opened 2018-08-22
[[ Bug 21514 ]] Automatically select text of fields in PI

bwmilby:
Add `openField` handler to select the text of fields. For the custom
properties behavior, when a key is selected in the tree view widget, then
the corresponding key text will also be selected.


montegoulding:
@bwmilby I think we have an issue now that we merged the tree view patch with a note for bug 20503. I think we will need to open a new report just to track the auto text selection.

bwmilby:
Filed new bug report, referenced in the original, updated note file name.

montegoulding:
@livecode-vulcan review ok 2a73480

Thanks @bwmilby !

livecode-vulcan:
💙 review by @montegoulding ok 2a73480c084518f06edfa4d59bb3e0b7bc8e16b0

PR #1993, opened 2018-08-27
[[Bugfix 21530]] Add more FFI examples to the Extending LiveCode Guide

andrewferguson:
I've updated the C FFI section of the "Extending LiveCode" guide to include a couple more examples of wrapping C functions (including an example of wrapping one of Apple's C-based APIs) and a guide to utilizing the already-written C/C++ functions that are available through LiveCode's foundation library (libfoundation).

### DIFF
--- a/Documentation/guides/Extending LiveCode.md
+++ b/Documentation/guides/Extending LiveCode.md
@@ -740,7 +740,27 @@ We could use:
 		binds to "atof"
 
 See the [Language Reference](https://github.com/livecode/livecode/blob/develop/docs/guides/LiveCode%20Builder%20Language%20Reference.md#the-c-binding-string) 
-for more information on C binding strings. 
+for more information on C binding strings.
+
+##### Other Binding Examples
+
+* Binding to the [strlen function](http://www.cplusplus.com/reference/cstring/strlen/) in the C standard library:
+
+		foreign handler C_Strlen(in pString as ZStringNative) returns CUInt \
+		binds to "c:strlen"
+This function can the be called from within LiveCode Builder as follows:
+
+		variable tLength as Integer
+		put C_Strlen("Hello World") into tLength --tLength will have a value of 11
+
+* Binding to the [CFDataGetBytePtr function](https://developer.apple.com/documentation/corefoundation/1543330-cfdatagetbyteptr?language=objc) within Apple's C-based Carbon framework for macOS:
+
+		foreign handler C_CFDataGetBytePtr(in theData as ObjcId) returns Pointer \
+		binds to "c:Carbon.framework>CFDataGetBytePtr"
+Note that the binding string for Apple's C-based frameworks is typically of the
+form "c:FrameworkBundleName.framework>functionName", where *FrameworkBundleName*
+is the name of the framwork (e.g: Carbon / CoreGraphics / etc...) and *functionName*
+is the name of the function to bind to.
 
 ##### Callbacks
 
--- a/Documentation/guides/Extending LiveCode.md
+++ b/Documentation/guides/Extending LiveCode.md
@@ -746,17 +746,17 @@ for more information on C binding strings.
 
 * Binding to the [strlen function](http://www.cplusplus.com/reference/cstring/strlen/) in the C standard library:
 
-		foreign handler C_Strlen(in pString as ZStringNative) returns CUInt \
-		binds to "c:strlen"
+      foreign handler C_Strlen(in pString as ZStringNative) returns CUInt \
+         binds to "c:strlen"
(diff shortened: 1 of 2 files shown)


livecodepanos:
@livecodeali do you think this can be rebased against `develop-9.0`? It would be good to have it in LC 9.0.2 rc-1 too

livecodeali:
@livecodepanos Yes, I don't see why not

Documentation/guides/Extending LiveCode.md
```diff
@@ -740,7 +740,27 @@ We could use:
...
+         binds to "c:strlen"
+This function can the be called from within LiveCode Builder as follows:
+
+      variable tLength as Integer
+      put C_Strlen("Hello World") into tLength --tLength will have a value of 11
+
+* Binding to the [CFDataGetBytePtr function](https://developer.apple.com/documentation/corefoundation/1543330-cfdatagetbyteptr?language=objc) within Apple's C-based Carbon framework for macOS:
+
+      foreign handler C_CFDataGetBytePtr(in theData as ObjcId) returns Pointer \
+         binds to "c:Carbon.framework>CFDataGetBytePtr"
```
montegoulding, line 759: @andrewferguson `CFDataGetBytePtr` is in `CoreFoundation`

Documentation/guides/Extending LiveCode.md
```diff
@@ -810,6 +830,77 @@ are available in the LiveCode source repository. This is a more
...
+This gives the following foreign declaration for the `MCStringDecode` function:
+
+      foreign handler MCStringDecode(in pData as Data, in pDataFormat as CUInt, in pExternalRep as CBool, \
+         out pString as String) returns CBool binds to "<builtin>"
+
+This foreign binding now allows for a simple string decoding function to be written in LCB:
+
+      unsafe handler StringFromUTF8Data(in pData as Data) returns String
+         variable tDecoded as String
+         MCStringDecode(pData, 4, false, tDecoded)
```
montegoulding, line 888: It might be good to declare a constant here so 4 doesn't seem like such a magic number

Documentation/guides/Extending LiveCode.md
```diff
@@ -810,6 +830,77 @@ are available in the LiveCode source repository. This is a more
...
+* the `is_external_rep` parameter exists for legacy reasons only, and takes a value `false`
+* the `r_string` parameter is the resulting decoded string. Unlike all the other parameters
+  it should be declared as an `out` parameter because it gives the decoded String
+
+The method to bind to a foundation library function is the same as binding to any C function,
+with two important distinctions:
+
+* the name of the function in the LCB declaration must exactly match the name of the function
+  in the foundation library
+* the binding string is simply `<builtin>`, with no class or function name specified
```
montegoulding, line 877: Is it worthwhile explaining that any of `<builtin>`, `MCStringDecode` and `c:MCStringDecode` will work as a binding string here? Perhaps it's worth saying that if you want to name your foreign handler something other than the name of the function then you can use a binding string.

Documentation/guides/Extending LiveCode.md
andrewferguson: OK, this is (to me at least) odd:
The binding to CFDataGetBytePtr is successful and appears to work as expected when the framework is one of:
Carbon, Foundation, CoreFoundation, CoreGraphics, CoreAudio, CoreVideo,...

I have a feeling that maybe LiveCode is just ignoring the framework name and binding to the correct function. It doesn't, however, bind if the specified framework doesn't exist (e.g: HappyTime.framework gives "unable to load foreign library").

Is this expected behavior?

Documentation/guides/Extending LiveCode.md
andrewferguson: I'll add in a note about `c:MCStringDecode`. I don't think it is necessary to explain `MCStringDecode` because this is something that is not specific to binding to the foundation library and should instead be explained with the general format of C binding strings (probably in the language reference).

Documentation/guides/Extending LiveCode.md
runrevmark: I suspect it works for all of those because all of those frameworks depend on CoreFoundation (and thus you can import the symbol from it, because it has already imported that symbol); OR, it is because where various things have actually lived in the OS has changed through the mac versions (mac has the notion of 'Umbrella' frameworks, which is a framework which encompasses many others) and so they are all seen as 'essentially the same thing' at the OS level.

In the binding code in the engine, we always dlopen() the specified library, and then dlsym from that handle (as far as I'm aware) - which is why it fails if you pass a library which doesn't actually exist.

PR #2007, opened 2018-10-17
[Bug 14079] Ensure one can set subscripts/superscripts more than once

livecodepanos:
Previously a non-zero `textShift` was applied to the selectedChunk only if the existing `textShift` was `not a number`. This worked fine in the first attempt where the the existing textShift was empty, but did not work afterwards if the `textShift` had been set to 0 (since 0 _is_ a number).

So we also need to check if the existing textShift is zero.


livecodeali:
@livecode-vulcan review ok 862c437

livecode-vulcan:
💙 review by @livecodeali ok 862c43748304f0535c20757427b7bd8cffd61965

PR #2013, opened 2018-11-30
Fix ide test failure

livecodepanos:
1. Get `the substacks of stack tLine` only if tLine is actually a stack.
2. Changed the name of the `testInputLibrary` substack of the `livecodeTestInterface` plugin (as well as all references to it) to  `testInputLib`, as there was already a stack with the same name in the test system which was causing the tests to fail because of stack name collision.


livecodeali:
@livecode-vulcan review ok 2c0e313

livecode-vulcan:
💙 review by @livecodeali ok 2c0e31370ba9777e6a736ff72ab049849a0955bd

PR #2015, opened 2018-12-04
[19277] Ensure Property Inspector always remembers its position

livecodepanos:
The stack `revPropertyInspector 1`, when opened for the first time, inherits its top from the top stack.

In `revIDEEnsureOnscreen`, there is the line:

(1)
```
if the effective top of stack pStackName < item 2 of tScreenRect then
      set the top of stack pStackName to (item 2 of tScreenRect + abs(the top of stack pStackName - the effective top of stack pStackName))
else if ...
```

If `the topstack` has a negative `top`, then the first time you open the `revPropertyInspector` stack, the `if` condition in 1 is true, so the `top` of this stack changes, and is saved later in `revIDESetPaletteRectPreference`.

However, if you close this stack and reopen it from the menubar, again the `if` condition in 1 is true, since `the effective top of stack pStackName` (=`revPropertyInspector`) is negative. But its `top` is now not negative, since it is retained from `revIDEGetPaletteRectPreference`.

This results in its `top` increasing every time you open it.

A solution to this would be either to set the `top` of the revPropertyInspector stack to the `abs()` of `the top of the topstack` (as this patch does), or change the `if` block in (1) to compare against the actual `top`, not the `effective top`.


montegoulding:
@livecodepanos I think there may be a deeper issue here so I'm taking a quick look.

montegoulding:
Ok so I have nailed down the issue to the effective rect not being updated when the screen is locked. 

The following code clearly demonstrates the issue:
```
on mouseUp
   local tRects
   put the rect of this stack into tRects
   lock screen
   set the top of this stack to the top of this stack + 100
   put return & the effective rect of this stack after tRects
   put return & the rect of this stack after tRects
   unlock screen
   put return & the effective rect of this stack after tRects
   put tRects
end mouseUp
```

So I think:
- this behavior needs to be documented. Need to check to see if it's the same on all platforms as I'm fairly sure macOS has some special handling of resizing so it is put off while the screen is locked. My tests were on macOS.
- the issue may/will impact other windows so it `revIDEEnsureOnscreen` needs fixing rather than the palette code. Perhaps something like this will suffice?:

```
constant kEstimatedWindowDecorationHeight = 22 --> might not be a good default cross platform???
local tStackEffectiveTop
if the lockScreen then
    put the top of stack pStackName - kEstimatedWindowDecorationHeight into tStackEffectiveTop
else
    put the effective top of stack pStackName into tStackEffectiveTop
end if

if tStackEffectiveTop < item 2 of tScreenRect then
      set the top of stack pStackName to (item 2 of tScreenRect + abs(the top of stack pStackName - tStackEffectiveTop))
else if
```

montegoulding:
I guess an alternative might be for `revIDEEnsureOnscreen` to force unlock the screen before using an effective rect prop:
```
local tLockScreen = 0
repeat while the lockScreen
    unlock screen
    add 1 to tLockScreen
end repeat
--> do effective stuff
repeat while tLockScreen is not 0
    lock screen
    subtract 1 from tLockScreen
end repeat
```

mwieder:
I can verify the problem on linux as well: the effective rect doesn't get updated while the screen is locked. My results from that test:
342,148,742,548
341,122,743,549
342,248,742,648
341,222,743,649

montegoulding:
Same on Windows too:
343,151,743,551
335,120,751,559
343,251,743,651
335,220,751,659

montegoulding:
Also based on these tests we have title heights of:
- Windows: 29
- macOS: 22
- Linux: 26

So we probably want the estimated height to be 29 to make sure it's ok on windows.

runrevmark:
So the issue here is that computing the effective rect uses API calls on all three platforms using the current window handle, and thus the pre-lock-screen rect (as lock screen stops the engine changing properties of the window until the final unlock screen - otherwise stacks would move within lock screen blocks!).

Now, on Mac and Windows I *think* the effective rect of a window can be computed without a window handle - by using API calls which just use the window attributes and not a window handle (indeed, I think there is code at least on Windows which does this already somewhere). However, on Linux I'm pretty sure this isn't possible because that requires interoperation with the window manager, and the window manager only operates on windows...

However, there is a simple 'hack' we can put in if lock screen is in effect which will fix the case of only the 'rect' attribute of the window changing - in view_platform_getwindowrect (which is per-platform - although this only affects Mac, Windows and Linux), if MClockscreen is non-zero, then fetch the window handle's content rect, and frame rect, compute the difference (frame rect will contain content rect), then apply those differences to the stack's current rect. (Note some care with logical vs physical co-ord mappings maybe required).

The above should be fine on all platforms - indeed, I think it is only the setting of blendLevel and rect properties of stacks (i.e. windows) which are deferred during lock screen... Any other window attributes cause a recreation of the window handle immediately (and even if not, at least the cases where effective rect works uniformly is almost all, rather than only when not under lock screen).

livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/6815

PR #2017, opened 2018-12-05
Ensure "ideNewControl" msg is sent when importing an image from the IDE

livecodepanos:


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -7221,6 +7221,9 @@ on revIDEImportControl pType, pFileName, pReferenced
    set the name of tCreatedControlID to the last item of pFileName
    unlock messages
    unlock screen
+   if not (pReferenced is true) then
+      ideMessageSend "ideNewControl", tCreatedControlID
```
montegoulding, line 7225: @livecodepanos There's a few things I'm noticing here:

1. I'm not sure why with the image creation we are doing:
```
put the long id of the last control into tCreatedControlID
```
after `revIDECreateObject` Instead of:
```
put the result into tCreatedControlID
```

2. There are cases where `pReferenced` is `true` and  `ideNewControl` will be sent by `revIDECreateObject` (referenced video) and cases where `pReferenced` is not true and `ideNewControl` will not be sent (audio).

3. `revIDECreateObject` does this so we probably need to do that in cases where it is sent here:

```
   ## Prevent passing of ideNewControl message for objects created on IDE stacks
   if revIDEObjectIsOnIDEStack(tCreatedControlID) is false then
      // AL-2015-04-08: [[ Bug 14822 ]] Ensure 'edited' status of stack is set
      revIDESetEdited the short name of this stack
      ideMessageSend "ideNewControl", tCreatedControlID
   end if
```

montegoulding: Oh... one other thing... in the case of imported images I think we could use `revIDECreateObject` then:
```
set the text of tCreatedControlID to url ("binfile:" & pFilename)
```
But if we need to send the message for other imported things here then maybe no need to change that.

montegoulding: @livecodepanos thinking about this a little bit more I can't help wondering if we should move the `import` cases into `revIDECreateObject`... say by adding an optional `pFilename` parameter. I think that might simplify maintenance as all object creation is done in one handler and only setting of extra properties due to it being an import operation happens here. Then the image case for example just becomes:
```
          if pReferenced is true then
              revIDECreateObject "com.livecode.interface.classic.image", tTargetStack, tCreatedControlLocation
          else
              revIDECreateObject "com.livecode.interface.classic.image", tTargetStack, tCreatedControlLocation, pFilename
          end if
          put the result into tCreatedControlID
         if pFileName ends with ".svg" then
             -- SVG files can't be referenced
             set the text of tCreatedControlID to drawingSvgCompileFile(pFileName)
          else if pReferenced is true then
              set the filename of tCreatedControlID to pFileName
          end if
          set the borderWidth of tCreatedControlID to 0
          select tCreatedControlID
```

PR #2021, opened 2018-12-19
Fix DataGrid cached control error

trevordevore:
If form controls were cached, fixed line height was true, and showing the vertical scrollbar was set to auto then the DataGrid would attempt to draw itself prior to generating the list of cached controls. This would result in an empty line in the list of controls to draw from. This empty line would then be used in handlers that loop through the controls. Instead of a valid control reference an empty value would be used and an error would be triggered.


livecodepanos:
@livecode-vulcan review ok b0cf1a4

livecode-vulcan:
💙 review by @livecodepanos ok b0cf1a44c1a4f3c84520c0df01a34d488c739349

PR #2029, opened 2019-01-31
Make some IDE stacks to come to the front rather than closing

livecodepanos:
if they are already open but not the top stack

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -8861,9 +8861,11 @@ end revIDEToggleShowIDEStacksInLists
 on revIDETogglePalette pPaletteName
    local tStackName
    put revIDEPaletteToStackName(pPaletteName) into tStackName
-
+   
    if the mode of stack tStackName is 0 or the visible of stack tStackName is false then
       revIDEOpenPalette pPaletteName
+   else if tStackName is not line 1 of the openStacks then
+      go card 1 of stack tStackName
    else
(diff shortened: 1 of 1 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -8861,9 +8861,11 @@ end revIDEToggleShowIDEStacksInLists
 on revIDETogglePalette pPaletteName
    local tStackName
    put revIDEPaletteToStackName(pPaletteName) into tStackName
-
+   
    if the mode of stack tStackName is 0 or the visible of stack tStackName is false then
       revIDEOpenPalette pPaletteName
+   else if tStackName is not line 1 of the openStacks then
+      go card 1 of stack tStackName
```
montegoulding, line 8868: @livecodepanos I'm not sure about this patch because the toggle is meant to open and close the IDE stacks, however, I can imagine (I'm pretty sure I've done it) using the toggle when the target stack is obscured by other stacks. This is really what the Window menu is meant to do but that doesn't list all IDE stacks.

I would suggest the following:
- update `revIDEStackIsIDEWindow` to include any modeless or toplevel (if there are any) IDE stacks so that these stacks appear in the Window menu
- implement a function `revIDEStackIsObscured` to determine if the user can't see the stack because it is obscured by the rects of other stacks.
- use `revIDESendWindowToFront` to raise the window in `revIDETogglePalette` if it is `among the lines of the openStacks` and `revIDEStackIsObscured`

PR #2030, opened 2019-02-01
Ensure stack name cannot be set to a number via the Project Browser

livecodepanos:


livecodeali:
@livecodepanos Since there is logic in the IDE library for what consitutes a valid stack name (plus a warning if the stack name begins with rev), the best thing would probably be to factor out that code into an IDE function which is called by the PB and in the IDE library - something like
```
function ideCheckStackName pName
-- Code from revIDEPropertySet 
end ideCheckStackName
```
which could return false if the stack name is not valid (or if they choose Cancel if the first three chars are `rev`), so that the caller can exit the handler if it returns false or continue with the name change otherwise.

livecodeali:
@livecode-vulcan review ok 0c4856b

livecode-vulcan:
💙 review by @livecodeali ok 0c4856b4aa1ab35a9fffc13c091e250129cc058f

PR #2033, opened 2019-02-08
[21555] Ensure mobileControlTarget returns the long ID of the target if the target is a DataGrid

livecodepanos:
On mobile, the DG2 has native scrollers automatically. However, in this case `mobileControlTarget` returned the ID of the native scroller. This was problematic because when leaving and then revisiting the card that has the DG2, the ID was changing.

This patch ensures that `mobileControlTarget` will always return the *long ID* of the target, if the target is a DG2. This ensures `mobileControlTarget` returns a unique ID that does not change.

### DIFF
--- a/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
+++ b/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
@@ -324,8 +324,8 @@ private command _CreateMobileScroller
          (_ControlType() is "table" and \
          the dgProps["show hscrollbar"] of me is not false) then
       if sScrollerId is empty or sScrollerId is not among the lines of mobileControls() then
-         mobileControlCreate "scroller"
-         put the result into sScrollerId
+         put the short name of me into sScrollerId
+         mobileControlCreate "scroller", sScrollerId
       end if
       mobileControlSet sScrollerId, "canBounce", "true"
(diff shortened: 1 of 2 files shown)


Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
```diff
@@ -324,8 +324,8 @@ private command _CreateMobileScroller
          (_ControlType() is "table" and \
          the dgProps["show hscrollbar"] of me is not false) then
       if sScrollerId is empty or sScrollerId is not among the lines of mobileControls() then
-         mobileControlCreate "scroller"
-         put the result into sScrollerId
+         put the short name of me into sScrollerId
+         mobileControlCreate "scroller", sScrollerId
```
livecodemichael, line 328: @livecodepanos Using the name as the scroller ID here will cause issues if there are two open data grids with the same name. It is unlikely, but I guess you could use the long id of the control. Or maybe check to see if the name is not already in `mobileControls()`

livecodepanos: Thanks @livecodemichael, good point.

I have updated the title of the PR as well

livecodemichael:
@livecode-vulcan review ok 3f67262

livecode-vulcan:
💙 review by @livecodemichael ok 3f672629669138f268bd9914e52a54fe1e7dc4b2

PR #2034, opened 2019-02-11
[21831] Ensure you can copy/cut/clear grouped controls from the PB

livecodepanos:


livecodeali:
OK, so here's something that's really confusing:
`revIDEObjectsAreSelectable` is used to determine whether you can cut/clear objects through the contextual menu at the moment. But how do you pop up a contextual menu if the object is not selectable? Also, `revIDEObjectsAreSelectable` is incorrect to say that its function is to
```
Determine if a list of objects are selectable using the select command
```
As you can group a button, turn selectGroupedControls off and do `select button 1` just fine.
Given that it is only used for contextual menus and is a guard against failure of `revIDESelectObjects` (which is how the implementations of the picked functions work - i.e. select and then perform action, I think we can potentially get rid of `revIDEObjectsAreSelectable`, or at least its use in the menu bar context menu stuff. Unless something somewhere else in the IDE relies on either of these...

montegoulding:
Hmm `revIDEObjectsAreSelectable` could just have the same owner check removed as long as `selectGroupedControls` is not changed before `revIDESelectObjects`. Then it still guards against un-open controls or selecting across stacks etc. Of course if we change the implementation of cut, copy and clear to not involve selection then we don't have any of these concerns although we would have to expose the serialised object format to script in order to implement cut and copy that way.

montegoulding:
Screen Shot 2019-03-27 at 10 17 21 am
This is what we need to guard against and I think this patch as is would throw errors in this case.

montegoulding:
We could also remove the selectGroupedControls check from it I think

PR #2036, opened 2019-02-14
[21414] Fix error when closing the SE and the Find window is still open

livecodepanos:


Toolset/palettes/script editor/behaviors/revsefindmaincardbehavior.livecodescript
```diff
@@ -33,7 +33,9 @@ end menuPick
 
 on closeCard
    historySave
-   send "revSEFindFinalize" to (getCaller())
+   if there is a stack getCaller() then
+      send "revSEFindFinalize" to (getCaller())
+   end if
```
livecodeali, line 40: Can we `put getCaller() into tCaller` or something so it doesn't have to compute it twice?

livecodeali: Actually I'm not sure I understand this... what is the caller? `getCaller` returns `the long id of...` so under what circumstances does it not exist? Or is this check essentially just 'if the caller is a stack then...'?

livecodepanos: @livecodeali in this case `getCaller()` is `stack "revNewScriptEditor 1" of stack "/Users/panos/Desktop/github/livecode/ide/Toolset/palettes/script editor/revscripteditor.8.rev"`, so if you first close the SE while the Find window is open the following happens:
- the SE will close the Find dialog
- the `closeCard` handler of stack `revSEFindMainCardBehavior` will be executed
- `send "revSEFindFinalize" to (getCaller())` will throw an error "cannot find stack", since stack `stack "revNewScriptEditor 1"` is already closed

livecodepanos: I'll add `put getCaller() into tCaller`

livecodepanos: hmm this throws the same the same error, because `the cCaller of the owner of me` evaluates to `stack "revNewScriptEditor 1" ..` which is closed, so `the long id of the cCaller ..` throws a "cannot find stack", so I had to slightly modify the `getCaller()` implementation too

montegoulding:
@livecode-vulcan review ok c4d95c7

livecode-vulcan:
💙 review by @montegoulding ok c4d95c7a5bb8d2e12ab94ed03c31a74a6d469b8b

PR #2038, opened 2019-02-26
[21856] Unload Plugin stack from memory

livecodepanos:
See https://quality.livecode.com/show_bug.cgi?id=21856 for a detailed description of the problem

### DIFF
--- a/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
+++ b/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
@@ -104,11 +104,13 @@ on ideOpenStack pTarget
       if tLine ends with ".livecode" or tLine ends with ".livecodescript" or \
             tLine ends with ".rev" then 
          put the substacks of stack tLine into tSubstacks
+         // Unload Plugin from memory
+         delete stack tLine
       end if
       if tLine is the filename of stack tTargetStack OR \
             tSubstacks contains tTargetStack then
          exit ideOpenStack
-      end if
+      end if      
    end repeat
    
    if tTargetStack is not the short name of me then
--- a/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
+++ b/Toolset/palettes/start center/revStartCenterBehavior.livecodescript
@@ -96,24 +96,9 @@ on ideOpenStack pTarget
    local tTargetStack
    put the short name of ideStackOfObject(pTarget) into tTargetStack
    
-   local tPlugins
-   put revIDEGetAllPlugins() into tPlugins
-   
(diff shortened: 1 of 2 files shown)


Toolset/palettes/start center/revStartCenterBehavior.livecodescript
```diff
@@ -104,11 +104,13 @@ on ideOpenStack pTarget
       if tLine ends with ".livecode" or tLine ends with ".livecodescript" or \
             tLine ends with ".rev" then 
          put the substacks of stack tLine into tSubstacks
+         // Unload Plugin from memory
+         delete stack tLine
```
trevordevore: @livecodepanos Will this change IDE behavior if you always delete the plugin stack, even if it was already in memory?

livecodepanos: @trevordevore hmm good point. If the stack is already in memory but not open, then I believe this change will not have any other unwanted consequences. 

But if the stack is already open, then this line will cause the stack to close, which is something we don't want to. I guess we should do something like:

```
if not (the short name of stack tLine is in the openstacks) then
   delete stack tLine
end if
```

trevordevore: @livecodepanos This is better. Ideally it wouldn't alter the state of the stack at all (if the stack is among the mainstacks it should remain among the mainstacks) but that isn't possible with existing syntax. Purging a stack from memory can affect script local variables. `the stacks` returns stack filenames that have open stacks but there is no equivalent for filenames of `mainstacks`.

runrevmark:
@livecodepanos (and @trevordevore so he can make sure I'm not talking rubbish!) Wouldn't this work:
```
if the filename of stack (the mainstack of pTarget) is among the lines of revIDEGetAllPlugins() then
  exit ...
end if
```
It would avoid loading any stacks into memory at all.

trevordevore:
@runrevmark - From what I can tell that code wouldn't load any additional stacks into memory. I didn't look to see if it accomplishes the same goal. 

As a side note, I was longing for a `stacks`-like property the other day that would return the filenames of all `mainstacks`.

livecodepanos:
Nice one @runrevmark . I will slightly modify this snippet, as `pTarget` is not necessarily a stack (might be a card) so `the mainstack of pTarget` will throw an error.

runrevmark:
@livecode-vulcan review ok 9beee9c

livecode-vulcan:
💙 review by @runrevmark ok 9beee9c970262fb72ae705c00f6b28a0feb98c68

PR #2048, opened 2019-03-29
[[ Bug ]] Fix android binaries when BFS

montegoulding:
This patch fixes the non native binaries path for android architectures other
than armv6. These architectures were previously unhandeld so the the bin folders
were not found.


livecodeali:
@livecode-vulcan review ok f838d74

livecode-vulcan:
💙 review by @livecodeali ok f838d748430307ff9f87e647c67c2aadaa916cd0

PR #2050, opened 2019-04-03
[[ Bug 21484 ]] Allow scroll wheel to work in PI text fields

livecodeali:
Include a `rawKeyDown` handler in the `revinspectoreditorbehavior` to
intercept pgup/pgdn/scroll wheel messages when inside a text field.  If
the field is not at either limit, then the messages will scroll the text
field.  If at a limit, then the messages will bubble up to the parent card.

Add `openField` handler to select the text of fields.  For the custom
properties behavior, when a key is selected in the tree view widget, then
the corresponding key text will also be selected.


montegoulding:
@livecode-vulcan review ok d1050c0

Would be good to squash merge these

livecode-vulcan:
💙 review by @montegoulding ok d1050c0469798ab339882700796c2ec46cc62473

PR #2063, opened 2019-05-13
[[ Bugfix MultipleLCBDeps ]] Added chained LCB dep checking

livecodemichael:
The IDE extension library has been updated to make sure that
the full dependency chain is checked. That if a wideget is
dependent on a module, we need to make sure that the modules
dependecies are included (before the module itself).


montegoulding:
@livecode-vulcan review ok 19a8a5a

livecode-vulcan:
💙 review by @montegoulding ok 19a8a5a07ed0ca40551d3759ab2bfc371ff90189

PR #2065, opened 2019-05-19
[[ FasterDG ]] Use container layerMode in form view

runrevmark:
This patch updates the DataGrid so that it takes advantage of container
layer mode when accelerated rendering is enabled for the stack.

The appropriate layer mode settings are now applied to the datagrid
group and its child groups - e.g. the row groups are all dynamic whilst
all ancestor groups are container.

To minimize unnecessary re-rerendering of the row templates, checks for
not needing to change the background color of the row template background
graphic and checks for not needing to refocus have been added.

A new property `minimal layout` has been added which, when set to true,
will only dispatch `LayoutControl` to a row-template if its data has
changed, or if its size has changed.

Currently the datagrid group must have its `showBorder` property set to
false to take advantage of accelerated rendering.

### DIFF
--- a/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
+++ b/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
@@ -239,6 +239,44 @@ private command _Initialize
       local theMasterRect
       put the rect of me into theMasterRect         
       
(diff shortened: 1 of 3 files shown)


livecodeali:
@livecode-vulcan review ok c5e8b24

livecode-vulcan:
💙 review by @livecodeali ok c5e8b245ce014bd75d7095982347debf8b348b80

PR #2066, opened 2019-05-21
Ensure Extensions Builder doesn't throw error about missing builtin modules

livecodepanos:


livecodeali:
Fixed by https://github.com/livecode/livecode/pull/7070 instead

PR #2067, opened 2019-05-28
[22091] Ensure selecting the same SVG icon does not cause the icon to disappear

livecodepanos:


livecodeali:
@livecodepanos I think this should be nested `if ... then`s so that the `close this stack` only happens if the target is the icon list - otherwise the stack will close when another part is clicked on that doesn't handle mouseUp

livecodepanos:
Good catch @livecodeali :) Will push an update now

livecodeali:
@livecode-vulcan review ok cd9736d

livecode-vulcan:
💙 review by @livecodeali ok cd9736d77050a97e91ea22bc162bcddba48da5c5

PR #2071, opened 2019-06-03
[[ Bug 22137 ]] Fix conditional breakpoint evaluation

montegoulding:
This patch ensures that conditional breakpoints that evaluate to false are reevaluated
if the same breakpoint is encountered again. For example, if a conditional breakpoint
is in a repeat loop it should be evaluated each iteration. A previous patch indended to
ensure that errors encountered while debugging a previous event were not re-entered
caused conditional breakpoints to not re-evaluate each iteration if no other breakpoint
was encountered.


livecodepanos:
@livecode-vulcan review ok db8f273

livecode-vulcan:
💙 review by @livecodepanos ok db8f27307da0e2e0844513bf0c06443dd0c0880c

PR #2077, opened 2019-08-08
[Bug 22285] Prevent PI custom props taking focus from SE

BerndN:


livecodepanos:
Thanks for the PR @BerndN.

I should have said to submit the PR against `develop-9.5` branch, not `develop`, as `develop-9.5` is now our main maintenance branch.

I have rebased the PR for you.

livecodepanos:
Hello Bernd, 

This patch does fix the problem with the script editor, but can throw an error in cases where `the selectedField` is empty. You have to `put true into grevdevelopment` to see these errors. 

For example if you open close the PI, open the SE and start moving the stack around, you'll get an error, because the custom props of the stack change (since you move the stack), so this handler is called, but `the selectedField` is empty in this case. If we just add an extra check:

```
if the selectedField is not empty then
      local tActiveField
      put revRuggedID(the long id of the selectedfield) into tActiveField
      if tActiveField contains not "revNewScriptEditor" then
            select the text of field "key" of me
      end if
end if
```

then this will prevent the "key" field of the custom props in the Property Inspector from being highlighted when clicking on the treeView widget.

 So I think it should be better to have something more general, like:

```
if the long name of the target contains "widget" && quote & "array" & quote \
               && "of group" && quote & "customProperties" & quote && "of group" && quote \
               & "Custom Properties" & quote && "of group" && quote & "inspector" & quote \ 
                  then
         select the text of field "key" of me
end if
```

mwieder:
See also https://quality.livecode.com/show_bug.cgi?id=22297.

BerndN:
@livecodepanos 
what do you think of 

```
         if the selectedField is  empty then
            select the text of field "key" of me
         else
            local tActiveField
            put revRuggedID(the long id of the selectedfield) into tActiveField
            if tActiveField contains not "revNewScriptEditor" then
               select the text of field "key" of me
            end if
         end if
```

it is more specific to the problem which is what you preferred.

livecodepanos:
@BerndN This ^^^ will throw an error (you have to `put true into grevdevelopment` first to see the error) if you click on the stack to grab it and move it around, just after closing the custom props Property Inspector:

Screen Shot 2019-08-08 at 21 07 58

I believe @mwieder is right and we have to fix the root of the problem, which is probably that "idePropertyChanged" messages are still sent to the property inspector after it has been closed.

I will ask the team to chime in and give some input.

BerndN:
@livecodepanos 
I can not trigger the debugger although I did "put true into grevdevelopment" following your recipe repeatedly.

Anyway the patch 2 I provided should not fire if the PI is closed.
After all it says "select the text of field "key" of me" of note "Of Me" 
who is "me" if the PI is gone? The script should not be triggered at all.

I agree that an explanation would be that the PI lingers somewhere may be in a parallel universe

mwieder:
@BerndN LOL. And the "who is me" question is exactly why the error is triggered. I can't submit a patch on github to fix the unsubscribe message because it's in a binary stack. But it's a simple fix and someone on the team should have no problem with it.

BerndN:
@mwieder 
I think the PI is pulled together from scripts.
The code pertaining to the stack script of the PI is in 

revInspectorTemplate.livecodescript

So you probably could do a pull request.
I just tested with "beep" in revInspectorTemplate and PI-2 did beep

mwieder:
Great. I always thought that what was missing in the PI was a beep <g>.
Thanks for the hint. I'll go poke around in there.

mwieder:
My patch submitted as https://github.com/livecode/livecode-ide/pull/2078.

BerndN:
@livecodepanos 
I changed the patch to your code. Your code also avoids the PI coming to the front if field "key" is highlighted and one changes a property of the inspected control and PI is not in front.

PR #2078, opened 2019-08-08
fix for bug report 22297

mwieder:
This patches the property inspector to do a revUnsubscribeAll on closing.

The problem reported in bug report 22297 being that the property inspector subscribes to the idePropertyChanged message but doesn't unsubscribe from that message if the selected object hasn't changed, as for example if you try to edit the script of an object after having looked at its properties. This will trigger a recurring error display and require closing the IDE.

### DIFF
--- a/Toolset/palettes/inspector/revinspectortemplate.livecodescript
+++ b/Toolset/palettes/inspector/revinspectortemplate.livecodescript
@@ -29,8 +29,8 @@ end openStack
 
 on closeStack
    ## Unsubscribe from all messages
-   revIDEUnsubscribe "idePreferenceChanged:idePropertyInspector_labels"
-   
+   revIDEUnsubscribeAll
+
    set the cSelectedObjects of me to empty
    
    # Unlock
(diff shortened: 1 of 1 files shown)


livecodepanos:
Hello all,

This patch - although it seems correct anyway - does not fix the original problem. An `idePropertyChanged` message is still sent to the Property Inspector when it is closed, and the SE throws an error. 

Recipe:
- in the msg box: `put true into gRevDevelopment`. Close msg box
- Create a stack
- Open the PI for the stack
- Go to custom properties
- Add a couple of custom props, and click on a custom prop, so as the "key" field is highlighted
- Close the Property Inspector
- Click on the "Code" button in the LC menubar to open the script editor for the stack

--> Error

livecodepanos:
Hmm I think I have found the problem.

The new line added in this patch, i.e. `revIDEUnsubscribeAll` with no param, unsubscribes the target from all subscribed messages. In the case of the recipe above ^^^^ the target is the *card* of the Property Inspector template stack. However, the Property inspector *stack* itself has not unsubscribed from all messages at this point. So doing this explicitly, by adding an extra line as below seems to fix the problem:

```
revIDEUnsubscribeAll -- Unsubscribes the card
// ADDED THE FOLLOWING LINE
revIDEUnsubscribeAll the long id of me -- Unsubscribes the stack
```

PR #2084, opened 2019-10-01
[Bug 22396] fix literal "/*" in Auto Complete If Structures causing e…

BerndN:
…ndless loop

Fix consists of emptying all quoted Strings

### DIFF
--- a/Toolset/palettes/script editor/behaviors/revsecommoneditorbehavior.livecodescript
+++ b/Toolset/palettes/script editor/behaviors/revsecommoneditorbehavior.livecodescript
@@ -2082,6 +2082,7 @@ function autoCompleteIFsAreBalanced pLineNumber, pScript
       if tCharBegin is 0 then exit repeat
       put offset(tEndAsterix, pScript) into tCharEnd
       if tCharEnd = 0 then exit repeat
+      if tCharBegin > tCharEnd then exit repeat
       delete char tCharBegin to tCharEnd of pScript
    end repeat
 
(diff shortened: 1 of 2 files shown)


mwieder:
@livecodepanos @BerndN - thanks.

Toolset/palettes/script editor/behaviors/revsecommoneditorbehavior.livecodescript
```diff
@@ -2051,6 +2051,26 @@ function autoCompleteIFsAreBalanced pLineNumber, pScript
...
    put line 1 to tCounter of pScript into pScript
+   
+   local tWorkText, tOldDelim
+
+   -- remove quoted literals
+   put the itemDelimiter into tOldDelim
+   set the itemDelimiter to quote
+   repeat for each line aLine in pScript
+      if aLine contains quote then
+         repeat with i = 1 to the number of items of aLine
```
mwieder, line 2062: I think this could be faster:
put 1 into i
repeat for each item tItem in aLine
 if i mod 2 = 0 then
  put empty into item i of aLine
 end if
 add 1 to i
end repeat

mwieder: ...that was supposed to be a multiline comment for lines 2062 through 2066. I guess I don't have this multiline thing down yet.

BerndN: As you and I discovered when timing the two versions there was no discernible difference.
Thank you for your suggestion.

PR #2085, opened 2019-10-01
Add a refresh button footer to refresh the view

mwieder:
The Project Browser annoyingly loses its content regularly. This patch adds a button to the toolbar to refresh the content when it gets messed up. I reused the cog icon because I couldn't figure out how to add a new one or where the current ones come from. Unlikely to be confusing though.

### DIFF
--- a/Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
+++ b/Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
@@ -140,7 +140,7 @@ private command doRefreshProjectView
    repeat for each line tStack in tStacks
       put the cardIDs of stack tStack into tCards
       repeat for each line tCardID in tCards
-         refreshChildren the long name of card id tCardID of stack tStack
+         refreshChildren the long id of card id tCardID of stack tStack
       end repeat
    end repeat
    
--- a/Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
+++ b/Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
@@ -133,7 +133,7 @@ on preOpenStack
 end preOpenStack
 
 on doRefreshProjectView
-    local tStacks, tCards
+    local tStacks, tCards, tControls
 
     lock screen
     put the openstacks into tStacks
@@ -142,6 +142,13 @@ on doRefreshProjectView
         repeat for each line tCardID in tCards
             refreshChildren the long name of card id tCardID of stack tStack
         end repeat
+    end repeat
+
+    put the controlIDs of card "list" of stack "revIDEprojectbrowser" into tControls
+    repeat for each line tGroup in tControls
+        if the short name of control id tGroup is "containerRow" then
+            set the visible of control id tGroup of card "list" of stack "revIDEprojectbrowser" to true
+        end if
     end repeat
     refreshProjectView
     unlock screen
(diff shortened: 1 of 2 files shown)


Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
```diff
@@ -76,7 +76,8 @@ on preOpenStack
    addFrameItem "groupControl","footer", "action", "Group controls", "folder open alt", "","groupControls", the long id of me
    addFrameItem "cloneControl","footer", "action", "Clone control", "files alt", "","cloneControls", the long id of me
    addFrameItem "deleteControl","footer", "action", "Delete control", "trash", "","deleteControls", the long id of me
-   
+   addFrameItem "refreshControl","footer", "action", "Refresh display", "cog", "","refreshProjectView", the long id of me
```
BerndN: how about "refresh"
addFrameItem "refreshControl","footer", "action", "Refresh display", "refresh", "","refreshProjectView", the long id of me

mwieder:
Thanks! That's the piece I was missing. Are these documented somewhere? Where are the icons stored?

BerndN:
they are the names of the svg icons, not sure about documentation

livecodepanos:
Thanks for the PR and the comment @mwieder and @BerndN.

This patch can be enhanced so as to take care of cases of grouping/ungrouping controls.

Use case:
- Open a stack and open the PB. Expand stack and card in the PB
- Add a couple of controls. Notice they do appear in the PB as expected.
- Select a couple of the controls you added and click on the Group button. Notice that an empty group appears in the PB. 

What should appear is a non-empty group, that will contain the controls you selected in the previous step.

So I suggest the line you added to call a new handler , say `doRefreshProjectView` instead of the existing `refreshProjectView`, which will refresh everything. Something like:
```
on doRefreshProjectView
   local tStacks
   put the openstacks into tStacks
   repeat for each line tStack in tStacks
      repeat with x=1 to the number of cards of stack tStack
         refreshChildren the long name of card x of stack tStack
      end repeat
   end repeat
   refreshProjectView
end doRefreshProjectView
```

mwieder:
Done, although I used a "repeat for each" loop and locked the screen, both in the interest of speed. I'd be even happier if grouping objects automatically refreshed the owning object, but I'll look into that as a separate pull request.

livecodepanos:
Thanks @mwieder :)

 Looks good to me. Pinging @livecodeali for a second look.

livecodepanos:
Ah @mwieder also there is bug report about it (https://quality.livecode.com/show_bug.cgi?id=17916), so if you could add a release note (`notes/bugfix-17916.md`) it would be great.

(Just make sure the note starts with a `#` followed by a space char)

Thanks!

mwieder:
Would still like to know where those little icons are stored. Can't find them anywhere.

livecodeali:
Check the `Icon SVG Library` docs in the dictionary. It was originally just fontAwesome hard-coded as SVG icons, essentially. @bwmilby added the ability to add icons to it.

https://github.com/livecode/livecode/blob/develop/extensions/libraries/iconsvg/iconsvg.lcb

mwieder:
The refresh button didn't actually bring the display back :-o so the doRefreshProjectView handler now properly displays the information. If this is a datagrid there may be a different way of doing this, but this works now.

Toolset/palettes/project browser/revprojectbrowserbehavior.livecodescript
```diff
@@ -132,6 +132,21 @@ on preOpenStack
...
 
+on doRefreshProjectView
+    local tStacks, tCards
+
+    lock screen
+    put the openstacks into tStacks
+    repeat for each line tStack in tStacks
+        put the cardIDs of stack tStack into tCards
+        repeat for each line tCardID in tCards
+            refreshChildren the long name of card id tCardID of stack tStack
```
BerndN: refreshChildren the long name of card id tCardID of stack tStack
should be 
refreshChildren the long id of card id tCardID of stack tStack

long name breaks e.g 
function ideStackOfObject pLongID
   local tOffset
   put wordOffset("stack",pLongID) into tOffset

if card name contains "Stacks in Use" as for the message box

mwieder:
Got something that actually works properly now. A brute force method is necessary when the display goes blank, and since it's an intermittent thing there's no way to set up a unit test. Still, some day I'd like to find out what causes the blank display instead of just fixing the problem after it occurs.

mwieder:
still waiting for a review after nine months?

PR #2089, opened 2019-10-03
Feature fast stack introspection

mwieder:
I finally got fed up with revIDEStacks() taking 49 seconds to run, so these patches at least bring it down to under ten seconds.

### DIFF
--- a/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/Toolset/libraries/revidelibrary.8.livecodescript
@@ -1109,15 +1109,11 @@ function __readStructureOfContainer @pData
     local tControls
 
     put the childControlIDs of the target into tControls
-    repeat for each line tCardID in tControls
-#   repeat with x = 1 to the number of controls of the target
+    repeat with x = 1 to the number of lines in tControls
       # read the properties of the control
       dispatch function "__readPropertiesOfControl" to control x of the target with sControlPropertiesToRead, pData[x]
       
-      --local tName
-      --put the name of control x of the target into tName
-      
-      #  if the control is a group
+       #  if the control is a group
       if the name of control x of the target begins with "group" then
          
          dispatch function "__readStructureOfContainer" to control x of the target with pData[x]["controls"]
@@ -1181,18 +1177,19 @@ private command __fetchPropertyCustomGetter pGetter, pObject, pProperty, @xPropA
    end if
 end __fetchPropertyCustomGetter
 
+# MDW 2019.10.02 [[ feature_fast_stack_introspection ]] added pProperties parameter for speed
 private on __fetchPropertyOfControl pProp, @xData
    local tError
-   put empty into tError
-   
    local tObjID
-   put the long id of the target into tObjID
+
+   put empty into tError
+   if pProperties is empty then
+      put the long id of the target into tObjID
+      put __classicObjectProperties(tObjID) into pProperties
+   end if
    
-   local tProperties
-   put __classicObjectProperties(tObjID) into tProperties
-   
-   if tProperties["properties"][pProp]["getter"] is not empty then
-      __fetchPropertyCustomGetter tProperties["properties"][pProp]["getter"], \
+   if pProperties["properties"][pProp]["getter"] is not empty then
+      __fetchPropertyCustomGetter pProperties["properties"][pProp]["getter"], \
             tObjId, pProp, xData
       exit __fetchPropertyOfControl
    end if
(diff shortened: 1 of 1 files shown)


Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -1181,18 +1177,19 @@ private command __fetchPropertyCustomGetter pGetter, pObject, pProperty, @xPropA
    end if
 end __fetchPropertyCustomGetter
 
+# MDW 2019.10.02 [[ feature_fast_stack_introspection ]] added pProperties parameter for speed
 private on __fetchPropertyOfControl pProp, @xData
```
BerndN: should'nt there be a third parameter "pProperties" for "__fetchPropertyOfControl"?

mwieder: Yep - you're commenting on an earlier commit. Somehow my latest edits didn't get in, so I edited again and pushed the corrected file as 82d41bb.

livecodeali:
@mwieder This is great! Have you tried using `card id tCardId` and `control id tControlId` instead of a numeric `card x`, `control x`? In theory it should be faster for most cases, at least when it has already been run once. Anyway it would be interesting to see the benchmarks.

mwieder:
@livecodeali I'm out of town for a week, but I'll check when I get back. I remember doing that and then realizing I needed the x counter anyway, but I don't offhand remember the details.

mwieder:
@livecodeali I'm back in town and now I remember why I couldn't use the card id. It's line 1123, "add tControlsProcessedInContainer to x" that prevents a sequential uses of each card id. That's also the reason for the trap at the end to exit the repeat loop instead of continuing after reaching the last card id.

livecodeali:
@mwieder Ah! That's not necessary with your changes (indeed it is in conflict with them) as the `childControlIds` does not recurse into groups. So I think you should be able to get rid of both the counter and the whole `tControlsProcessedInContainer` thing and just iterate over lines of `the childControlIds`. Alternatively you could use `controlIds` instead of `childControlIds`, although personally I think the code is better as a whole without the counting (and is _hopefully_ more efficient to boot)

mwieder:
@livecodeali OK - I'll give it a try again. I originally used the cardIDs and ended up with too many iterations because of the "add to x" thing. So in that case returning the number of controls in tControls instead of returning tControlsProcessedInContainer should give the correct result?

mwieder:
Interesting. Benchmarking that change shows it to be twice as slow. Scratching my head as to why. Here's what I've got for the revised function (still need the counter for the array index):

function __readStructureOfContainer @pData
   local tControlsProcessedInContainer
   
   # 1) Repeat for each control in the container
   #  MDW 2019.10.02 [[ feature_fast_stack_introspection ]]
   local x, tControls
   
   put the childControlIDs of the target into tControls
   put 1 into x
   repeat for each line tControlID in tControls
      # read the properties of the control
      dispatch function "__readPropertiesOfControl" to control id tControlID of the target with sControlPropertiesToRead, pData[x]
      
      #  if the control is a group
      if the name of control id tControlID of the target begins with "group" then
         
         dispatch function "__readStructureOfContainer" to control id tControlID of the target with pData[x]["controls"]
         put the result into tControlsProcessedInContainer
      end if
      add 1 to x
   end repeat
   
   return the number of lines of tControls
end __readStructureOfContainer

livecodeali:
@mwieder Twice as slow as what? The PR as it stands? One of the reasons the PR as it stands is quicker is that it will be exiting the repeat loop too soon due to adding the number of nested child controls to the index into the parent `childControlIDs`

Are you testing several implementations in parallel and ensuring that the array they return is the same for a given stack?

mwieder:
Well, that was twice as slow as the previous PR code, but you're right in that the array extents were correct but the contents were wrong. I've got it down to 16 seconds now from 48, but the childControlIds thing wasn't working out. Iterating over the childControlIds of the target was giving the wrong results, so I couldn't eliminate the counter. It would certainly be more elegant if that worked out, but I've spent most of the day fiddling with it and I can't seem to optimize this any farther.

mwieder:
@livecodeali *controlIDs*? that property seems to be recognized for cards and groups, but it isn't documented. From experiment when applied against a card it provides all the controls on the card, although it's rather dramatic when there's a datagrid on the card. When applied against a group it shows the childControlIDs plus the id of the group itself. It seems less than useful for this case because it would require examining the type of control being examined before diving in to gather the information.

At any rate, the internal functions are now three times faster than the original.

PR #2092, opened 2019-10-15
[[ Bug 22211 ]] Remove unnecessary 'unlock cursor' in ideMouseMove

livecodepanos:
This patch removes a seemingly unnecessary `unlock cursor` in the IDE's
mouseMove event handler (`ideMouseMove`).

The IDE intercepts mouseMove so that it can update the resize cursor
used when the mouse is over the relevant selection handlers. It does this
by locking the cursor in this case and then changing the cursor
appropriately; unlocking the cursor when there is a selected object and
the mouse is not over a selection handle.

Previously it would also unlock the cursor if pointer tool was in effect and
there was no selected object which (because the handler is executed in
time after any user handlers) would undo any user use of `lock cursor`.

Closes https://quality.livecode.com/show_bug.cgi?id=22211

### DIFF
--- a/notes/bugfix-22211.md
+++ b/notes/bugfix-22211.md
@@ -1,0 +1,1 @@
+# Ensure lock cursor does not break when the pointer tool is selected
--- a/Toolset/libraries/revbackscriptlibrary.livecodescript
+++ b/Toolset/libraries/revbackscriptlibrary.livecodescript
@@ -3335,8 +3335,6 @@ on ideMouseMove pX, pY, pTarget
       else
          unlock cursor
       end if
-   else
-      unlock cursor
    end if
 end ideMouseMove
 


runrevmark:
I suggest updating the title of the commit message to:
```
[[ Bug 22211 ]] Remove unnecessary 'unlock cursor' in ideMouseMove
```
And then have a commit message body:
```
This patch removes a seemingly unnecessary `unlock cursor` in the IDE's
mouseMove event handler (`ideMouseMove`).

The IDE intercepts mouseMove so that it can update the resize cursor
used when the mouse is over the relevant selection handlers. It does this
by locking the cursor in this case and then changing the cursor
appropriately; unlocking the cursor when there is a selected object and
the mouse is not over a selection handle.

Previously it would also unlock the cursor if pointer tool was in effect and
there was no selected object which (because the handler is executed in
time after any user handlers) would undo any user use of `lock cursor`.
```

notes/bugfix-22211.md
```diff
@@ -0,0 +1 @@
+# Ensure lock cursor does not break when the pointer tool is selected
```
runrevmark, line 1: I suggest
```
# Fix IDE interference with locking the cursor when pointer tool is selected

runrevmark (changes requested):

livecodepanos:
Setting the to WIP for further investigation, as this patch introduces this regression:

Recipe: Move from a selection handle horizontally left out of the selected objects rect.

Expected result: The cursor to switch from the resizing cursor to the usual one
Observed result: Cursor is stuck to resize cursor, even after leaving the selection handles of the selectedObject

livecodepanos:
Replaced by https://github.com/livecode/livecode-ide/pull/2132

PR #2104, opened 2019-11-07
[Bug 16933] Adapt icon size to text size in Project Browser

BerndN:


livecodepanos:
Thanks @BerndN for this.

I just thought that we could reduce code duplication by putting everything in a command in `revprojectbrowserbehavior.livecodescript` file, something like:

```
on resizeRowConnector pControl, pWhichRow
   local tTextSize, tWidgetSize
   put the effective textsize of pWhichRow into tTextSize
   switch 
      case tTextSize >= 8 and tTextSize < 12
         put 11 into tWidgetSize
         break
      case tTextSize >= 12 and tTextSize < 14
         put 13 into tWidgetSize
         break
      case tTextSize >= 14 and tTextSize < 18
         put 15 into tWidgetSize
         break
      case tTextSize >= 18
         put 17 into tWidgetSize
         break
      default
         put 13 into tWidgetSize
   end switch
   
   set the width of widget "icon" of group pControl of pWhichRow to tWidgetSize
end resizeRowConnector
```

and then in each of these 5 *rowBehavior scripts just call:

`resizeRowConnector pControl,me`

Note I have not tested this yet.

BerndN:
@livecodepanos I agree that it is desirable to reduce code duplication. But those 5 behaviors are essentially already code duplication and the handler I changed 
command connectorResize pControl
is exactly the  same in all 5 behaviors. To make a handler 
resizeRowConnector pControl,me
and leave the rest in place does not really address code duplication in those 5 behaviors.

I would rather leave it as suggested and a code clean up could be done later for e.g. all  handlers 
command connectorResize pControl
in all 5 behaviors.
What do you think?

livecodepanos:
@BerndN ok sure, let's leave this patch as it is and we can do a cleanup at a later stage.

@livecode-vulcan review ok de4107d

livecode-vulcan:
💙 review by @livecodepanos ok de4107d516bd0a5f609e5390708e0d3835f8544b

PR #2117, opened 2020-01-31
[Bug 22555] Keep compile errors in SE w/o LiveError on

BerndN:
If LiveError is off compile errors are not flagged if one clicks in field "script" or any other action in SE.
Bug 22555


mwieder:
I don't understand this. If the code in both branches of the conditional is the same, why duplicate it? Why not just get rid of the 'if LiveErrors then' test?

BerndN:
This was the place I found where it did what I wanted. I am not even sure it is the best place to fix this. I wanted to change as little as possible as far as the basic structure is concerned and make it easy for the reviewer to review this patch.
On second thought I agree with you, it would be cleaner and functionally equivalent.

Maybe @montegoulding could shed some light on this

PR #2133, opened 2020-05-13
[[ Bug 22285 ]] Restrict selection of key field in customprops editor

livecodeali:
This patch ensures that the key field of the custom properties
editor in the property inspector is only selected in response to
a change in the tree view hilite.

This is essentially just a tweak of #2077 but using the fact that `hiliteChanged` only happens when the hilite really changes.


runrevmark (approved):

runrevmark:
@livecode-vulcan review ok 47f395e

livecode-vulcan:
💙 review by @runrevmark ok 47f395e295038cb569f4ae233bc27bb26bc73694

PR #2136, opened 2020-06-25
Bugfix 22798

macMikey:
mobile datagrid scrollers are initialized with
delayTouches = true
this will prevent a datagrid from scrolling because the moment the datagrid is touched, a row will be selected.

### DIFF
--- a/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
+++ b/Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
@@ -368,7 +368,7 @@ private command _CreateMobileScroller
       mobileControlSet sScrollerId, "canBounce", "true"
       mobileControlSet sScrollerId, "pagingEnabled", "false"
       mobileControlSet sScrollerId, "canScrollToTop", "true"
-      mobileControlSet sScrollerId, "delayTouches", "false"
+      mobileControlSet sScrollerId, "delayTouches", "true"
       mobileControlSet sScrollerId, "canCancelTouches", "true"
       __PollVisibility
    end if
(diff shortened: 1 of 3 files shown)


Toolset/palettes/revdatagridlibrary/behaviorsdatagridbuttonbehavior.livecodescript
```diff
@@ -368,7 +368,7 @@ private command _CreateMobileScroller
       mobileControlSet sScrollerId, "canBounce", "true"
       mobileControlSet sScrollerId, "pagingEnabled", "false"
       mobileControlSet sScrollerId, "canScrollToTop", "true"
-      mobileControlSet sScrollerId, "delayTouches", "false"
+      mobileControlSet sScrollerId, "delayTouches", "true"
```
mwieder, line 371: Should this rather be merged into develop-9.6 to up its priority/urgency?

macMikey: Well, I didn't see a 9.6 branch open and
nobody else commented on the how list and
nobody else commented on the bug report and
this has been like this since whenever dg2 came out but nobody else seemed to notice, so I'm going with "maybe".
I mean, I personally think it should be a hotfix and get a special code name and a bug bounty Bond car to me and a lifetime free subscription to every LC premium thing, but I'm also megalomaniac.

MWCoastMedia: I agree with Mikey: this bug is driving me nuts.

mwieder: Don't hold your breath waiting for a pull request to be reviewed.
Meanwhile, this is a stack-only script, so you can always patch your own copy.

PR #2141, opened 2020-08-21
[Bugfix-22873] - Complete dgEditMode description

livecodealice:
Added missing text to the entry for dgEditMode and removed a reference that was appearing blank in the Dictionary.

### DIFF
--- a/Documentation/dictionary/datagrid.lcdoc
+++ b/Documentation/dictionary/datagrid.lcdoc
@@ -2658,14 +2658,14 @@ Description:
 
 Use the <dgEditMode> property to put the data grid into or take it out of edit
 mode. When in edit mode, the data grid displays an action control of the left
-hand side of each row (as defined by ) and a reorder control on the right. The
-appearance and behavior of these controls can be customized.
+hand side of each row (as defined by <edit mode action control>) and a reorder 
+control on the right. The appearance and behavior of these controls can be 
+customized.
 
 Only data grids form type data grids with <fixed row height> can be put into
 edit mode.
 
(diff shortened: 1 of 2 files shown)


Documentation/dictionary/datagrid.lcdoc
```diff
@@ -2658,14 +2658,14 @@ Description:
...
+hand side of each row (as defined by <edit mode action control>) and a reorder 
+control on the right. The appearance and behavior of these controls can be 
+customized.
 
 Only data grids form type data grids with <fixed row height> can be put into
 edit mode.
 
-Related: edit mode action control (property), edit mode action select control
-(property), edit mode reorder control (property),
+Related: edit mode action control (property), edit mode reorder control (property),
```
livecodepanos: The reference for `edit mode action select control` does not show because there is a line break between `edit mode action select control` and `(property)`. I suggest fixing this by doing:

```
Related: edit mode action control (property), 
edit mode action select control (property), edit mode reorder control (property),
```

There is a similar problem with:
 ```
EditModeHideActionControl
 (command), 
```
 (i.e. the line breaks, so the reference to this command is not shown). Could you address this issue as well?

livecodepanos: @livecodesam the reference to `edit mode action select control (property)` is still missing

livecodepanos:
@livecode-vulcan review ok a62fd92

livecode-vulcan:
💙 review by @livecodepanos ok a62fd92c065031cc533db8afba861a7bbc96be75

PR #2151, opened 2021-01-28
[[ Bug 22300 ]] Ensure right-clicking on the PB selects the correct row

livecodepanos:
This patch fixes an issue where under some circumstances an incorrect object row was selected when right-clicking on the Project Browser. This happened because `itemOffset` was called without `wholeMatches=true`, returning an incorrect value, e.g. `itemOffset("1","5,3,2,13,6,1")` returned `4` instead of `6`, thus an incorrect row was selected.


livecodeali:
@livecode-vulcan review ok e7e6917

livecode-vulcan:
💙 review by @livecodeali ok e7e6917593fe1086d1b227e8ad6513fc41081f8e

PR #2161, opened 2021-07-08
[[ Bug 23257 ]] Fix build error when "Allow Http Connections" is checked

livecodepanos:
This patch fixes an android build error when "Allow Http Connections" is checked.
This happened because the button "Allow Http Connections" was part of the group "permissions",
thus the group scripts affected the button under specific circumstances. This patch places the
"Allow Http Connections" button outside the "permissions" group, so as it is not affected by code
that applies to the group.


livecodeian:
@livecode-vulcan review ok 0f209f4

livecode-vulcan:
💙 review by @livecodeian ok 0f209f4e0dfa241362a0c36c972851e2ddd9f456

